PR #43, opened 2014-07-02
MarkerCluster, OMS and SearchBox integration

jqnatividad:
added OverlappingMarkerSpiderfier and integrated with SearchBox;
upgraded MarkerClustererPlus from 2.0.9 to 2.1.2


PeterTheOne:
- cluster properties were added in #140
- markerclustererplus update is done in also #166 

Spiderfier still would be nice.

netbrain:
LGTM

JeroenDeDauw:
This definitely needs rebase against master. And preferably gets turned into smaller, easy to review and merge, commits.

I guess this is quite dead at this point though. Will close if there is no further activity here in the next month or two.

PR #54, opened 2014-10-06
Reduced size of map JSON for performance reasons

JeroenDeDauw:


PeterTheOne:
Should we fix this PR or close it? I think the intention behind this is to make the json code smaller.

The has file moved to `/src/Maps/Elements/Location.php`. Almost the same thing could be done with:

    $array = array_filter( $array, function( $key ) use( $array ) {
        return !isEmpty( $array[$key] );
    } );

JeroenDeDauw:
I do not recall what the intent here was.

@HelmutLeitner can you comment on this?

HelmutLeitner:
@JeroenDeDauw, all my code changes were done to speed up the code, in effect to support thousands of markers instead of hundreds (within an answer time of a few seconds). 

As far as I remember, after speeding up data base access, the ammount of data that runs through the data interface, sent to the client, came on the time critical path. It was possible to remove some unneeded elements, and reduce the request time again by 20-30%, just by removing elements like "variable=0", but this was not possible with all such elements. The openstreetmap interface is neither efficient nor completely logical.

The code change PeterTheOne suggested is cleaner and more beautiful but probably slower. It means to put data elements into an (rather slow, dynamical memory managed, hashed) array to later remove again. Optimization often means to write more code, and even ugly code, to avoid such inefficiencies. 

It may also be that array_filter does not yield equivalent array elements. I do not know. 
Any change should be done measuring the effects on a 10000 marker test system, and exact timing. Only the "stop watch" is right.

PeterTheOne:
I merged master into this Pull Request and made a small fix and tests pass now. But I didn't comprehensively test it. Should we merge it?

src/Maps/Elements/Location.php
```diff
@@ -308,13 +308,26 @@ public function getJSONObject( $defText = '', $defTitle = '', $defIconUrl = '',
...
-			'address' => $this->getAddress(),
+			// not used for now:
+			//'alt' => 0,
 			'icon' => $this->hasIcon() ? \MapsMapper::getFileUrl( $this->getIcon() ) : $defIconUrl,
-			'group' => $this->hasGroup() ?  $this->getGroup() : $defGroup,
-			'inlineLabel' => $this->hasInlineLabel() ? $this->getInlineLabel() : $defInlineLabel,
-			'visitedicon' => $this->hasVisitedIcon() ? $this->getVisitedIcon() : $defVisitedIcon,
 		];
+		$val = $this->getAddress();
+		if( !empty( $val ) ) {
```
JeroenDeDauw: Need to use strict checks. `empty` will return true for string "0".

src/Maps/Elements/Location.php
```diff
@@ -308,13 +308,26 @@ public function getJSONObject( $defText = '', $defTitle = '', $defIconUrl = '',
 		$array = [
 			'lat' => $this->coordinates->getLatitude(),
 			'lon' => $this->coordinates->getLongitude(),
-			'alt' => 0,
-			'address' => $this->getAddress(),
+			// not used for now:
+			//'alt' => 0,
```
JeroenDeDauw: Can just remove this comment

JeroenDeDauw:
Thanks. The empty check needs to be fixed before merge.

PeterTheOne:
Fixed.

PR #59, opened 2015-01-04
Restore precision after breaking change in DV Geo

JeroenDeDauw:


JeroenDeDauw:
@thiemowmde FYI

thiemowmde:
Uh, wow, this is odd. Really odd. Why does "4, 2" become "4, 4"? Do you have an idea which commit in Geo did this?

JeroenDeDauw:
https://github.com/DataValues/Geo/commit/351d8c0c9084d5e16e61d98f42d88b76b31a184d

Which was part of a minor release.

JeroenDeDauw:
Users will now get `4° 0' 0.0000", 2° 0' 0.0000"` instead of `4° 0' 0", 2° 0' 0"`, though there appears to be little I can do about that here without hiding the decimal places altogether.

PR #71, opened 2015-03-21
OpenLayers/OSM fixes

sarumpaet:
1. Using protocol relative URLs fixes browser warnings if Wiki is running via HTTPS. Unfortunately opencyclemap doesn't seem to support HTTPS (properly) yet.

2. Markers in OpenLayers didn't handle touchstart events. Markers couldn't get selected.


JeroenDeDauw:
Awesome, thanks!

PR #105, opened 2016-02-06
Add ext.maps.services

mwjames:
To keep internals of a service hidden and only expose an interface so
that external users have access to it via the RL.

```
var service = new maps.services( $( this ) );
service.render( 'google' );
```

refs https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1404#issuecomment-180504474

### DIFF
--- a/Maps.resources.php
+++ b/Maps.resources.php
@@ -83,6 +83,22 @@ return call_user_func( function() {
 				'jquery.ui.dialog',
 			),
 		),
+
+		'ext.maps.services' => $moduleTemplate + array(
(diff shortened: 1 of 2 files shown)


mwjames:
I now can simply call the service without me having to know Maps internals (classes, function names etc.).

```
		var semanticNearBy = new nearBy(
			$( this ),
			new mw.Api(),
			new maps.services( $( this ) )
		);

		...

		).done( function ( ) {
			self.maps.render( self.parameters.maps );
		} );
```

PR #111, opened 2016-03-01
add MapQuest Layer

krabina:


JeroenDeDauw:
Belongs together with #112

I'm not able to test this. Would be nice to get confirmation from another user that this works as expected

krabina:
a working example can be seen here: http://bacid.kdz.eu/

JeroenDeDauw:
Thanks!

JeroenDeDauw:
Can you also submit an update to the release notes?

PR #117, opened 2016-03-25
expose marker functions and map

PeterTheOne:
This change makes it possible to add Listeners to the map and to add or remove markers from the outside.


JeroenDeDauw:
Thanks!

JeroenDeDauw:
Looks good. Might require breaking release. Waiting with merge based on https://github.com/SemanticMediaWiki/SemanticMaps/pull/52

PeterTheOne:
I found a problem with this: the now exposed properties are global and not per instance as it should be.

PeterTheOne:
There needs to be a different way to extend jquery.googlemap.js, jquery.leaflet.js, etc. from SemanticMaps..

JeroenDeDauw:
Now you say so... how did I not see that? :) Guess I've been out of the JS fun for too long.

The design of the Maps JS code is quite foobar (or indeed, missing altogether) (with the PHP being not much better). I'm thinking that responsibilities should be untangled and dependency injection be used. If you provide something that works without addressing the design issues that is entirely fine. It might be easier to get certain stuff done after solving some though.

JeroenDeDauw:
I unfortunately do not really have the time to check, but it's possible what @mwjames did in https://github.com/JeroenDeDauw/Maps/commit/e1a5d4c5b9b2de44eff378bed7a48ec17bfb07df is relevant

PeterTheOne:
Thanks!

PeterTheOne:
> I unfortunately do not really have the time to check, but it's possible what @mwjames did in e1a5d4c is relevant

I had a look at it, but it doesn't help in this case.

Instead I went for a global list of initialized maps. This still has the problem that you need to wait (setTimeout) for the maps to initialize and the list to be set.

includes/services/Leaflet/jquery.leaflet.js
```diff
@@ -39,7 +39,7 @@
 				icon:icon
 			};
 
-			var marker = L.marker( [properties.lat, properties.lon], markerOptions ).addTo( this.map );
+			var marker = L.marker( [properties.lat, properties.lon], markerOptions ).addTo( this.mapp );
```
JeroenDeDauw: huh?

includes/services/OpenLayers/jquery.openlayers.js
```diff
@@ -92,8 +94,8 @@
 				}
 
 				if (bounds != null) bounds.extend(location.lonlat); // Extend the bounds when no center is set.
-				var marker = this.getOLMarker(curLayer, location);
-				this.markers.push({
+				var marker = _this.getOLMarker(curLayer, location);
+				_this.markers.push({
```
JeroenDeDauw: Perhaps I'm being a bit dense here, but I'm not seeing where `_this` is defined

JeroenDeDauw:
So the main thing this does now is putting all map instances into `window.maps.mapList`, correct? How do you intend to use this? The instances in there can be of different types, which by the looks of it do not have a consistent public interface.

includes/services/Leaflet/jquery.leaflet.js
PeterTheOne: Ups, I should've double checked.. fixed now.

includes/services/OpenLayers/jquery.openlayers.js
PeterTheOne: It is defined a few lines down, but I reverted the change because it's unnecessary.

PeterTheOne:
> So the main thing this does now is putting all map instances into `window.maps.mapList`, correct?

correct. I use it like this: https://github.com/SemanticMediaWiki/SemanticMaps/pull/52/files#diff-57d942e3b9f63f91e79d69b9995b791aR56

>  The instances in there can be of different types, which by the looks of it do not have a consistent public interface.

You are right, this is a problem.

PeterTheOne:
For the SemanticMaps ajax funcitonality I need googlemap, leaflet and openlayers to implement `removeMarkers()` and `addMarker(properties)` (`removeMarker(marker)` would be nice too) or an `addMarker(properties)` that doesn't add Markers that already exist.

JeroenDeDauw:
Sounds reasonable. I wonder if it's not better to expose the real map objects rather than forcing people to go through the wrappers created by Maps

PeterTheOne:
Via the wrapper one can already access the real map object, which is good.

includes/ext.maps.common.js
```diff
@@ -14,5 +14,8 @@ window.maps = new ( function( $, mw ) {
 			mw.log( message );
 		}
 	};
-	
-} )( jQuery, mediaWiki );
\ No newline at end of file
+
+	this.googlemapList = [];
```
JeroenDeDauw: odd this is not `googlemapsList`

JeroenDeDauw:
Looks good. Do you want this to be merged in its current state, before you finished the work on the code that uses it in Semantic Maps?

includes/ext.maps.common.js
PeterTheOne: done in 94c6c5a

PeterTheOne:
Yes, please merge it in the current state. There might be changes in the future but it's stable for now.

PR #121, opened 2016-04-01
Add position to wgResourceModules array (required for MW 1.26+)

kghbln:


Maps.resources.php
```diff
@@ -16,6 +16,7 @@
 	$pathParts = ( explode( DIRECTORY_SEPARATOR . 'extensions' . DIRECTORY_SEPARATOR, __DIR__, 2 ) );
 
 	$moduleTemplate = array(
+		'position' => 'top',
```
kghbln: This is a more complex setup than usual. However I guess this is the right spot since it applies to all modules using the template.

JeroenDeDauw:
Not sure about if this needs top or not

PR #124, opened 2016-04-02
3.6 alpha: Require MW 1.23+ and PHP 5.5+

JeroenDeDauw:


kghbln:
Shouldn't this be a 4.0? Not a 100% semver match but pretty breaking.

JeroenDeDauw:
> *What should I do if I update my own dependencies without changing the public API?*

> That would be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package should have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

-- http://semver.org/

So no, this does not require bumping to 4.0.

kghbln:
That's what I meant with "Not a 100% semver match" :) I just figured since PHP and MW ... No worries. :)

PR #130, opened 2016-04-06
[DO NOT MERGE] Add support for extension.json

### LINKED ISSUE #126: Add support for MWs new extension registration
This needs to be done in such a way the extension still works with older versions of MW (down to MW 1.23) and continues to work via Composer in the way existing users expect it to.

Some of the features of the extension registration should also not be used:
- Hooks
- autoloading
- SpecialPages, Actions, JobClasses, APIModules

This one is up for the taking. Help with it is much appreciated.

paladox:
Fixes #126


JeroenDeDauw:
Err... looks like you where working based on some older version of the extension or something like that

paladox:
@JeroenDeDauw yep sorry, I rebased my fork for master and I'm re working it so it uses the new version, sorry about that.

PR #131, opened 2016-04-06
Add support for extension.json

### LINKED ISSUE #126: Add support for MWs new extension registration

paladox:
Fixes #126


JeroenDeDauw:
Needs rebase

paladox:
Ok rebased.

Probaly needs some testing by you since you would know what to expect please.

paladox:
Im not sure why it fails with https://travis-ci.org/JeroenDeDauw/Maps/jobs/135067644

JeroenDeDauw:
Thanks for working on this.

It fails because this code did not get executed

```
 	$GLOBALS['wgParamDefinitions']['coordinate'] = [
 		'string-parser' => 'DataValues\Geo\Parsers\GeoCoordinateParser',
 		'string-parser' => 'DataValues\Geo\Parsers\GeoCoordinateParser',
 	];
```

I probably should have specified this more clearly in the ticket: I do not want references to PHP constructs, be it class names, or function names, in the new extension registration thing. IMO it's a bad idea to support those there, and I'm sad it got implemented like that in MediaWiki.

paladox:
Oh how should add it. Should we do it through a callback or the extension function.

Do you have an example on how you would like it done please.

JeroenDeDauw:
I don't care so much where this is put, as long as it is in PHP space, and it works. It's easy enough to move it around anyway. Btw, I recommend using `::class` to reference class names without using strings. So instead of

```
	$GLOBALS['wgParamDefinitions']['coordinate'] = [
		'string-parser' => '\DataValues\Geo\Parsers\GeoCoordinateParser',
	];
```

You'd have

```
	$GLOBALS['wgParamDefinitions']['coordinate'] = [
		'string-parser' => \DataValues\Geo\Parsers\GeoCoordinateParser::class,
	];
```

Like this things don't break if someone renamed the class with a proper tool.

JeroenDeDauw:
Needs rebase again

@paladox is this something you want to continue with?

kghbln:
Ping @paladox

kghbln:
@paladox No worries if you do not like to continue on this. Please just say so if this is the case ...

paladox:
Hi, please completely forgot about this. I had a tab open for this but something else came up and then I forgot about this.

kghbln:
No worries. I know such situations very well. So you are continuing on this?

paladox:
I rebased now.

Maps.php
```diff
@@ -109,40 +127,19 @@
...
+
+	$GLOBALS['wgHooks']['ParserFirstCallInit'][] = 'MapsHooks::onParserFirstCallInit3';
+
+	$GLOBALS['wgHooks']['ParserFirstCallInit'][] = 'MapsHooks::onParserFirstCallInit4';
+
+	$GLOBALS['wgHooks']['ParserFirstCallInit'][] = 'MapsHooks::onParserFirstCallInit5';
+
+	$GLOBALS['wgHooks']['ParserFirstCallInit'][] = 'MapsHooks::onParserFirstCallInit6';
+
+	$GLOBALS['wgHooks']['ParserFirstCallInit'][] = 'MapsHooks::onParserFirstCallInit7';
```
JeroenDeDauw: Not sure why you did this replace. I like the older form much more.

extension.json
```diff
@@ -0,0 +1,344 @@
...
+		],
+		"ParserFirstCallInit": [
+			"MapsHooks::onParserFirstCallInit1",
+			"MapsHooks::onParserFirstCallInit2",
+			"MapsHooks::onParserFirstCallInit3",
+			"MapsHooks::onParserFirstCallInit4",
+			"MapsHooks::onParserFirstCallInit5",
+			"MapsHooks::onParserFirstCallInit6",
+			"MapsHooks::onParserFirstCallInit7"
+		],
```
JeroenDeDauw: This is now both here and in `Maps.php`

extension.json
```diff
@@ -0,0 +1,344 @@
...
+			},
+			"distance": {
+				"string-parser": "Maps\\DistanceParser::class"
+			},
+			"wmsoverlay": {
+				"string-parser": "Maps\\WmsOverlayParser::class"
+			},
+			"mapsimageoverlay": {
+				"string-parser": "Maps\\ImageOverlayParser::class"
+			},
```
JeroenDeDauw: As I wrote before in the thread, please leave this out of the JSON file

Maps.hooks.php
```diff
@@ -10,6 +10,123 @@
...
+			throw new Exception( 'You need to have Validator installed in order to use Maps' );
+		}
+
+		$GLOBALS['egMapsStyleVersion'] = $GLOBALS['wgStyleVersion'] . '-' . Maps_VERSION;
+	}
+
+	public static function onExtensionFunction() {
+		if ( $GLOBALS['egMapsDisableExtension'] ) {
+			return true;
+		}
```
JeroenDeDauw: This is now duplicated with code in `Maps.php`

JeroenDeDauw:
Something was broken with these changes as shown by the tests: https://travis-ci.org/JeroenDeDauw/Maps/jobs/238720483

Maps.php
paladox: Needs to be done otherwise we would have duplicate code and would need to update it in two places.

extension.json
paladox: Yes, to keep combat with mw 1.23.

Maps.hooks.php
paladox: Yep has to be done since users can bypass Maps.php and load extension.json directly.

extension.json
paladox: Thanks, done

paladox:
Im not sure how this change breaks the tests.

extension.json
JeroenDeDauw: Why not have this only in `Maps.php`? I don't see any advantage of having it in the JSON file

Maps.hooks.php
JeroenDeDauw: Sounds like this again is for compat reasons. You can drop support for MW < 1.27.

JeroenDeDauw: Hah, I just realized that MW support for 1.23 ended 2 days ago anyway

extension.json
paladox: As we can remove support for 1.27 and lower we can now keep this in extension.json. Also users can overide Maps.php by loading extension.json only. php entry points are also deprecated.

Maps.hooks.php
paladox: Yep

PR #135, opened 2016-04-11
Consistently switch to use __DIR__

kghbln:


JeroenDeDauw:
Thanks. However please wait for review instead of self merging

kghbln:
> However please wait for review instead of self merging

I was just testing how to delete branches. Apart from that: You're right!

Issue #138, opened 2016-05-17
Icons clustering makers not shown


waandershz:
For some days now the icons for clustering markers aren't shown anymore.
See: https://www.semantic-mediawiki.org/wiki/Maps_examples/Google_Maps_marker_clustering

We use them here: https://www.deltaexpertise.nl/wiki/index.php/Dijkvakken_Zeeland_VN

Has something to do with references Google-sites. Console browser gives error messages like 
> Failed to load resource: the server responded with a status of 404: https://google-maps-utility-library-v3.googlecode.com/svn/trunk/markerclustererplus/images/m1.png

Also see:
- http://stackoverflow.com/questions/37087583/google-maps-not-working-in-https
- https://github.com/googlemaps/js-marker-clusterer/issues/55
- https://wp-types.com/forums/topic/404-error-in-toolset-maps-due-to-outdated-googlecode-refs/

PeterTheOne:
https://github.com/JeroenDeDauw/Maps/blob/637e120d81681ed71646ddeb2e85cb9be2ef06d4/includes/services/GoogleMaps3/gm3-util-library/markerclusterer.js#L1578 should point to a local copy of the images..

waandershz:
@PeterTheOne So? We just use the extension and this code:

```
{{#ask: [[Category:Dijkvak]]
| ?Coordinate
| limit=500
| markercluster=yes
| format=map
}}
```

kghbln:
@waandershz There is nothing you can do on wiki.

waandershz:
@kghbln What do you mean? We just have to wait till next extension version?

kghbln:
Affirmative.

PR #139, opened 2016-05-17
Update link to markercluster resource

kghbln:
Fix location of markercluster files. Refs issue https://github.com/JeroenDeDauw/Maps/issues/138

* New code location according to: [stackoverflow - 37179980](http://stackoverflow.com/questions/37179980/markercluster-v3-stopped-working-properly?answertab=votes#tab-top)
* Code link set according to: [stackoverflow - 17341122](http://stackoverflow.com/questions/17341122/link-and-execute-external-javascript-file-hosted-on-github/18049842#18049842)

### DIFF
--- a/includes/services/GoogleMaps3/gm3-util-library/markerclusterer.js
+++ b/includes/services/GoogleMaps3/gm3-util-library/markerclusterer.js
@@ -1575,7 +1575,7 @@ MarkerClusterer.BATCH_SIZE_IE = 500;
  * @type {string}
  * @constant
  */
-MarkerClusterer.IMAGE_PATH = "//google-maps-utility-library-v3.googlecode.com/svn/trunk/markerclustererplus/images/m";
+MarkerClusterer.IMAGE_PATH = "//cdn.rawgit.com/googlemaps/js-marker-clusterer/gh-pages/images/m";
 
 
 /**


PeterTheOne:
The `m*.png` images https://github.com/googlemaps/js-marker-clusterer/tree/2f659a3c288cdb8dfd1940b00080ff31863b08bc/images should be included in Maps, because GitHub is not a [CDN](https://en.wikipedia.org/wiki/Content_delivery_network) and we may have the same problem in the future if we link to it.

PeterTheOne:
Actually we should not change the constant but pass the local imagePath in https://github.com/JeroenDeDauw/Maps/blob/637e120d81681ed71646ddeb2e85cb9be2ef06d4/includes/services/GoogleMaps3/jquery.googlemap.js#L620 as seen in https://github.com/googlemaps/js-marker-clusterer/blob/2f659a3c288cdb8dfd1940b00080ff31863b08bc/src/markerclusterer.js#L122 .

kghbln:
This would be a follow up commit I believe.

JeroenDeDauw:
So stuff can break if they change what is on the `gh-pages` branch of their repo? Seems like a bad idea if there is no stability guarantee

kghbln:
> So stuff can break

That's the situation we had during the past four years. So the path constant should be removed and the files provided by Maps?

PR #140, opened 2016-05-19
Add googlemaps3 clusterer properties

PeterTheOne:
Adds the following googlemaps3 clusterer properties:

    'gridSize': (number) The grid size of a cluster in pixels.
    'maxZoom': (number) The maximum zoom level that a marker can be part of a
        cluster.
    'zoomOnClick': (boolean) Whether the default behaviour of clicking on a
        cluster is to zoom into it.
    'averageCenter': (boolean) Wether the center of each cluster should be
        the average of all markers in the cluster.
    'minimumClusterSize': (number) The minimum number of markers to be in a
        cluster before the markers are hidden and a count
        is shown.

 – https://github.com/googlemaps/js-marker-clusterer/blob/gh-pages/src/markerclusterer.js#L41

They are all prefixed with "cluster" as in "clustergridsize", the default values are taken from [js-marker-clusterer](https://github.com/googlemaps/js-marker-clusterer) itself. Except clustermaxzoom that was set to null, has a default of 20. Before this change `averageCenter` was set to `true` statically and now is set to `false`, so this is a change of default behavior.

Usage:

    {{
    #display_map:0,0;1,1;1,0;0,1;5,5;6,6;5,6;2,0;2,1;-5,-5;-5,5;5,-5;-1,-1;-1,0;0,-1;10,10;-10,10;50,50;-50,50;50,-50;-50,-50;0,50;0,51;0,52
    | markercluster = true
    | clustergridsize = 120
    | clustermaxzoom = 20
    | clusterzoomonclick = true
    | clusteraveragecenter = true
    | clusterminsize = 2
    }}


JeroenDeDauw:
This is really cool, thanks for contributing. I'm wondering about the naming of the parameters. Perhaps separating the words with underscores is nicer for the user? Or perhaps just the `cluster` prefix? Not sure if underscore is the recommended separator. Thoughts @kghbln @mwjames?

kghbln:
I do not like underscores at all so ... I think these parameters are fine as they are. Probalby `clusterminimumclustersize` could be shortened to `clusterminsize` to avoid a "Getüm" while still keep the purpose intuitively comprehensible . ;)

PeterTheOne:
> Probalby clusterminimumclustersize could be shortened to clusterminsize

I agree, so I changed it in 9990fd46

PeterTheOne:
Should we set the default of `averageCenter` to `true` again, what do you think?

kghbln:
> Should we set the default of averageCenter to true again, what do you think?

I guess since it was already in there like this it should stay "true" for the wikis to avoid changing their current setup.

kghbln:
Since you are introducing new system messages you need to include them into "/i18n/en.json" and "/i18n.qqq.json", too. An example qqq entry will be
```"maps-googlemaps3-par-clustergridsize": "{{maps-par|googlemaps3|clustergridsize}}",```

PeterTheOne:
done and done.

I used the https://github.com/googlemaps/js-marker-clusterer/blob/gh-pages/src/markerclusterer.js#L41 description for i18n.

kghbln:
Great effort. Thanks! @JeroenDeDauw The i18n is fine with me.

JeroenDeDauw:
Thanks for the input.

Needs rebase against master, otherwise seems good to go.

One thing you can add is an update of the release notes, describing that is new. Adding some examples to the wiki would also be great, though that is not possible until we deploy the new version.

kghbln:
> One thing you can add is an update of the release notes, describing that is new.

Yeah, could have come up with this, too. Let's talk about coolness. :)

> Adding some examples to the wiki would also be great, though that is not possible until we deploy the new version.

Indeed, a way could also be to add the examples to sandbox.smw.o (daily updates) and once I updated smw.o I could just copy the pages over. To cut it short: A similar structure for the examples will probably [be nice](https://www.semantic-mediawiki.org/wiki/Category:Maps_examples).

@PeterTheOne Not sure if you already have accounts at smw.o and sandbox.smw.o If not I will accept respective requests fast.

PeterTheOne:
Ok, I did a rebase. Sure I can create some examples.

kghbln:
Great, I just updated sandbox.smw.o to master.

PeterTheOne:
@kghbln http://sandbox.semantic-mediawiki.org/wiki/Google_Maps_marker_clustering

PR #155, opened 2016-05-26
Use address as title when there is an address and no title

JeroenDeDauw:


JeroenDeDauw:
Ugh. Boom. Not being able to run the tests locally is kinda annoying. No time to look into this now tho. Probably its just integration tests that need adjustment

PeterTheOne:
> Probably its just integration tests that need adjustment.

Yes.

tests/phpunit/parsers/LocationParserTest.php
```diff
@@ -134,4 +134,22 @@ protected function getInstance() {
 		return new LocationParser();
 	}
 
+	public function testGivenAddressAndNoTitle_addressIsSetAsTitle() {
+		$location = ( new LocationParser() )->parse( 'Tempelhofer Ufer 42' );
+
+		$this->assertSame( 'Tempelhofer Ufer 42', $location->getTitle );
```
JeroenDeDauw: What the heck. I forgot the function call brackets? :D

JeroenDeDauw:
I cannot easily finish this as I'm not able to run the tests locally. So if you'd like to see this happen, feel free to pick this up.

PeterTheOne:
I did some work on the unit tests but

    <display_map searchmarkers="all">
    New York City
    Brussels
    Berlin
    London
    Norway
    </display_map>

doesn't should the address as titles on the map for me.

JeroenDeDauw:
It's not nice to have tests bind to LocationParser, esp if the production code they test does not directly bind to it.

How about adding an alternate constructor to Location? Something like `Location::newTitledLocation( $latitude, $longitude )`

Not that many tests should be affected by this change in LocationParser anyway. If they are, that suggests a problem somewhere. Perhaps some should just assert lat and long, and not care about the rest of the Location. I'm not sure as I have not really looked at the involved code.

PeterTheOne:
> How about adding an alternate constructor to Location? Something like Location::newTitledLocation( $latitude, $longitude )

I did this in 38b79f1.

> Not that many tests should be affected by this change in LocationParser anyway. If they are, that suggests a problem somewhere. Perhaps some should just assert lat and long, and not care about the rest of the Location. I'm not sure as I have not really looked at the involved code.

Would be the cleanest solution, but I don't want to change https://github.com/JeroenDeDauw/Maps/blob/master/tests/phpunit/parserhooks/ParserHookTest.php#L88 anytime soon.

JeroenDeDauw:
Right, for that code I'd just use this new alternative constructor

JeroenDeDauw:
Pffft, this Location class is foobar. Even though it was way worse before the last refactor I did to it. No reason for this thing to be mutable. It should just take all its stuff in its constructor. And not have an alternative one that does geocoding based on global state services >_>

JeroenDeDauw:
Random rant about something else aside, this PR looks good now.

I'm a bit worried that this behavior change might not be desired for some people though. What if you want to have a map with a bunch of markers but no pop-ups? When using addresses, all of a sudden after updating Maps, you'll get pop-ups. And you won't be able to turn it off. (Or do we have a general off setting for those things?) Thoughts @kghbln?

Therefore I do not want to merge this just yet, as then It'd be released in a few days. I'd like to ask the users for feedback first. This could be done as part of the upcoming release mail.

kghbln:
> What if you want to have a map with a bunch of markers but no pop-ups? When using addresses, all of a sudden after updating Maps, you'll get pop-ups. And you won't be able to turn it off.

In this case it would be preferable to have a query parameter allowing to say yes or no.

PeterTheOne:
Should we name it (frame it) like `hidepopups` or `useaddressastitle`?

JeroenDeDauw:
Good question. Btw, I suspect the maps docs are not using the word popup, but rather something like "info windows". Should probably be consistent with the stuff users actually see.

PeterTheOne:
Ok, I'm in favor of something similar to `hidepopups`..

kghbln:
What about just `popups` with "yes/no" or (`addresstitle`) with "yes/no". Hmm..., prefer the latter since `popups` seems to be too generic. `hidepopups` or `useaddressastitle` aleady contain sort of a "yes" or "no"  indicating of the default setting which only a few other options to "Maps" do. Since we also have a `title` option this appears to be more consistent to me.

PeterTheOne:
One month later.. so how should we proceed. :P

kghbln:
@JeroenDeDauw Ping :P

JeroenDeDauw:
Having a `useaddressastitle` parameter is fine by me

includes/parsers/LocationParser.php
```diff
@@ -33,6 +43,8 @@ class LocationParser extends StringValueParser {
 	public function stringParse( $value ) {
 		$separator = '~';
 
+		$useaddressastitle = $this->getOption('useaddressastitle');
```
JeroenDeDauw: `$useAddressAsTitle = $this->getOption( 'useaddressastitle' );`

JeroenDeDauw:
Thanks!

kghbln:
Just comes to my mind: Shouldn't we also get an entry for `useaddressastitle` to show via the `#mapsdoc` parserfunction? I guess this option applies to all services.

PeterTheOne:
For now this only added an `useaddressastitle` option for the LocationParser, correct. I'm not sure where I would set this in the maps parser function.

kghbln:
I thought this is going to be an option to the "display_map" parser function. In this case it should probably go to the [services files](https://github.com/JeroenDeDauw/Maps/tree/master/includes/services) like "Maps_Leaflet.php" etc. like the other options. I may be wrong though. @JeroenDeDauw ?

JeroenDeDauw:
If it is not specific to mapping services, then it should not be done there.

Adding an argument to the displaymap parser function can be done in `getParameterInfo` of `MapsDisplayMap`. If you want the parameter to also be available in ask queries, then you can stuff it in `getCommonMapParams` there. Though perhaps it's better to just add it in `SMMapPrinter`.

PR #161, opened 2016-06-07
add removeMarkers to leaflet and openlayers

PeterTheOne:
add removeMarkers functions to leaflet and openlayers (googlemaps3 already has this function) as noted in #117.

This is required for https://github.com/SemanticMediaWiki/SemanticMaps/pull/52 to work with leaflet and openlayers.

### DIFF
--- a/includes/services/OpenLayers/jquery.openlayers.js
+++ b/includes/services/OpenLayers/jquery.openlayers.js
@@ -4,13 +4,15 @@
  *
  * @author Jeroen De Dauw <jeroendedauw at gmail dot com>
  * @author Daniel Werner
+ * @author Peter Grassberger < petertheone@gmail.com >
  *
  * @todo This whole JS is very blown up and could use some quality refactoring.
  */
 
-(function ($, mw) {
+(function ($, mw, OpenLayers) {
 	$.fn.openlayers = function (mapElementId, options) {
 
+		this.map = null;
 		this.options = options;
 
 		OpenLayers._getScriptLocation = function() {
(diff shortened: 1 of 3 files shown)


includes/services/OpenLayers/jquery.openlayers.js
```diff
@@ -64,45 +66,56 @@
...
+				data:markerData
+			});
+			curLayer.addMarker(marker); // Create and add the marker.
+		};
+
+		this.removeMarkers = function () {
+			var map = this.map;
+			$.each(this.groupLayers, function(index, layer) {
+				map.removeLayer(layer);
+			});
```
JeroenDeDauw: Does this not also remove things that are not makers? Such as images and whatnot? And "base layers"? Perhaps the later are not removed, and perhaps the former should be counted as images (or rather as what this function is actually about). I'm very fuzzy on the stuff, though someone should check.

JeroenDeDauw:
Generally looks good to me. Did you test that the changes don't cause obvious breakage? (Yay for not having any tests that cover this stuff at all)

includes/services/OpenLayers/jquery.openlayers.js
PeterTheOne: First I wanted to remove markers directly but, some layers also need to be removed. The `removeMarkers` function only removes `groupLayers` they are only used for markers. Lines, etc. have their own layers like `lineLayer`.

PeterTheOne:
I did test my use case and it did not break, but I haven't tested with shapes like lines, etc.

PR #164, opened 2016-06-07
Update Leaflet to version 0.7.7

### LINKED ISSUE #160: Update Leaflet version to 0.7.7 or 1.0.*
Now we use version 0.4.5 from 2012, we should update to 0.7.7 or 1.0.*

CHANGELOG: https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md

PeterTheOne:
fix #160

As far as I can see we didn't use anything from the breaking changes: https://github.com/Leaflet/Leaflet/blob/master/CHANGELOG.md

My simple use case of displaying some markers looks good. Testing wanted. ^^


JeroenDeDauw:
You added `leaflet-src.js` but left `leaflet.js` unchanged?

PeterTheOne:
I replaced `leaflet.js`, it just isn't shown because the diff ist too big.. ;)

You I remove `leaflet-src.js`?

JeroenDeDauw:
> Should I remove leaflet-src.js?

If it is not used, then there is no point in having it in the repo no?

PeterTheOne:
It is sometimes useful to include it for debugging but I'm open to removing it.

JeroenDeDauw:
Does not matter all that much

Issue #67, opened 2015-03-13
Merge "missing" system messages


kghbln:
The following system messages are used by both [googlemaps](https://semantic-mediawiki.org/wiki/Help:Google_Maps_format), [openlayers](https://semantic-mediawiki.org/wiki/Help:OpenLayers_format) and [leaflet](https://www.semantic-mediawiki.org/wiki/Help:Leaflet_format) should be "aliased":

**maps-par-zoom**
* [x] maps-googlemaps3-par-zoom -> maps-par-zoom
* [x] maps-openlayers-par-zoom -> maps-par-zoom
* [x] maps-leaflet-par-zoom -> maps-par-zoom

**maps-par-searchmarkers**
* [x] maps-googlemaps3-par-searchmarkers -> maps-par-searchmarkers
* [x] maps-openlayers-par-searchmarkers -> maps-par-searchmarkers

**maps-par-resizable**
* [x] maps-googlemaps3-par-resizable -> maps-par-resizable
* [x] maps-leaflet-par-resizable -> maps-par-resizable
* [x] maps-openlayers-par-resizable -> _already_ maps-par-resizable

JeroenDeDauw:
Should we keep this open?

kghbln:
Yes. I think it is just a matter of replacing the dupes?

kghbln:
@JeroenDeDauw Finally did this. To much pressure right now by @PeterTheOne to make the next release even more fluffier. :)

PR #167, opened 2016-06-11
Provide missing system messages by merging parameter messages

kghbln:
These system messages were common to all mapping services and now merged into one system message (also removed identical thus duplicate system messages):
* maps-par-zoom
* maps-par-searchmarkers
* maps-par-resizable

Refs issue https://github.com/JeroenDeDauw/Maps/issues/67

Do formatting tweaks while I was at it


JeroenDeDauw:
If there where no different zoom messages, then this is fine. I do recall that zoom values differ between mapping services, with some using bigger numbers for more zoomed in, and some using them for more zoomed out.

kghbln:
> If there where no different zoom messages

No it basically tells that one can set the zoom level without further info.

Issue #141, opened 2016-05-19
Add clustering options for leaflet


PeterTheOne:
* openlayers seems to have some build in functionality.
* leaflet has a plugin https://github.com/Leaflet/Leaflet.markercluster

also see: #140

PeterTheOne:
Maybe before that update Leaflet #160 and OpenLayers #88.

PeterTheOne:
Leaflet is now up to date.

PeterTheOne:
I will wait for Leaflet `1.0.x` stable, update to that and then add https://github.com/Leaflet/Leaflet.markercluster (Which also has `spiderfy`-options)

PR #174, opened 2016-06-20
Update to leaflet version 1.0.0(-rc2)

PeterTheOne:
I think we shouldn't merge this yet, let's wait for 1.0.0. But we can use this PR for testing.

Also see: #141


kghbln:
Getting close to -rc2 https://github.com/Leaflet/Leaflet/issues/4732

PeterTheOne:
This branch is now on [leaflet 1.0.0-rc2](http://leafletjs.com/2016/07/16/leaflet-1.0-rc2.html). I did have a quick look if it works and it did but I haven't tested comprehensively.

PeterTheOne:
I'm ok with merging this before 1.0.0 is out, if anyone wants to. But we could also wait for the last bugs to be fixed.

kghbln:
> I'm ok with merging this before 1.0.0 is out

Me too since it would allow testing on sandbox before the next release of Maps gets out to the world. @JeroenDeDauw Thoughts or action? :)

PR #176, opened 2016-06-21
Removed broken layers functionality

### LINKED ISSUE #168: Disable or remove Layer: functionality
Regarding the functionality tied to the Layer: namespace introduced in Maps 3.0 https://github.com/JeroenDeDauw/Maps/blob/master/RELEASE-NOTES.md#new-features

This code was added in a dubious state and has received no maintenance since. I am not aware of anyone using this functionality, or documentation of it. I would thus not be surprised if it is currently broken.

Some serious issues with the functionality have been brought to my attention. These issues require follow up. If no one presents a good argument for keeping this functionality, I'm inclined to remove it.

If anyone is aware of users or documentation of this functionality, or is otherwise awesome of its status, please point this out.

JeroenDeDauw:
Fixes #168


JeroenDeDauw:
Huh, nothing blew up, even without me running the tests? That just makes me think stuff did blow up but the tests did not catch it :)

JeroenDeDauw:
This was about 1/3rd of the PHP code in Maps...

JeroenDeDauw:
@kghbln if the sandbox totally breaks now, you know who to throw the tomatoes at

PeterTheOne:
Indeed a lot of code.. O.o

kghbln:
> if the sandbox totally breaks now, you know who to throw the tomatoes at

Just to make sure. Should these tomatoes be rotten? :D

kghbln:
I does not. Just updated to latest master and at least we do not have fatals. Error log does not seem to collect issue information either.

JeroenDeDauw:
Thanks @kghbln

PR #180, opened 2016-07-04
add GoogleMaps3 API key

### LINKED ISSUE #179: Add support for Google Maps key
Google Maps now requires a key for new sites created after 2016-06-22 as reported [here](https://www.mediawiki.org/wiki/Topic:T71rtmkqrkuz4ig8).

Quote: 
```
As of 22 June 2016, Google Maps require a key for new sites, i.e. sites first using Google Maps after that date. See https://developers.google.com/maps/pricing-and-plans/standard-plan-2016-update 

Are there any plans to update the Maps extension? 

In the meantime, you can get an API key for free from https://console.developers.google.com/flows/enableapi?apiid=maps_backend,geocoding_backend,directions_backend,distance_matrix_backend,elevation_backend&keyType=CLIENT_SIDE&reusekey=true 

You can then edit the file "\extensions\Maps\includes\services\GoogleMaps3\Maps_GoogleMaps3.php". In the function "getApiScript", change the line: 

"return Html::linkedScript( '//maps.googleapis.com/maps/api/js?' . wfArrayToCgi( $urlArgs ) );"

to: 

"return Html::linkedScript( '//maps.googleapis.com/maps/api/js?key={YOURKEY}&v=3.24?' . wfArrayToCgi( $urlArgs ) );", replacing {YOURKEY} with the API key that was generated for you at the Google site.

By the way, Jeroen, thank you for all your good work.

Henrik (www.irhb.org)
```

PeterTheOne:
fixes #179

Haven't tested this yet.

### DIFF
--- a/includes/geocoders/Maps_GoogleGeocoder.php
+++ b/includes/geocoders/Maps_GoogleGeocoder.php
@@ -37,9 +37,11 @@ final class MapsGoogleGeocoder extends \Maps\Geocoder {
 	protected function getRequestUrl( $address ) {
 		$urlArgs = [
 			'address' => urlencode( $address ),
-			'sensor' => false,
-			'key' => $GLOBALS['egMapsGMaps3ApiKey']
+			'sensor' => false
 		];
+		if ( !empty( $GLOBALS['egMapsGMaps3ApiKey'] ) ) {
+			$urlArgs['key'] = $GLOBALS['egMapsGMaps3ApiKey']
+		}
 
 		return 'http://maps.googleapis.com/maps/api/geocode/xml?' . wfArrayToCgi($urlArgs);
 	}
--- a/includes/geocoders/Maps_GoogleGeocoder.php
+++ b/includes/geocoders/Maps_GoogleGeocoder.php
@@ -35,7 +35,13 @@ final class MapsGoogleGeocoder extends \Maps\Geocoder {
 	 * @return string
 	 */	
 	protected function getRequestUrl( $address ) {
-		return 'http://maps.googleapis.com/maps/api/geocode/xml?address=' . urlencode( $address ) . '&sensor=false';
+		$urlArgs = [
+			'address' => urlencode( $address ),
+			'sensor' => false,
+			'key' => $GLOBALS['egMapsGMaps3ApiKey']
+		];
+
+		return 'http://maps.googleapis.com/maps/api/geocode/xml?' . wfArrayToCgi($urlArgs);
 	}
 	
 	/**
--- a/includes/geocoders/Maps_GoogleGeocoder.php
+++ b/includes/geocoders/Maps_GoogleGeocoder.php
@@ -40,7 +40,7 @@ final class MapsGoogleGeocoder extends \Maps\Geocoder {
 			'sensor' => false
 		];
 		if ( !empty( $GLOBALS['egMapsGMaps3ApiKey'] ) ) {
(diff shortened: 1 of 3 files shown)


kghbln:
@PeterTheOne [FYI](https://www.mediawiki.org/w/index.php?title=Topic:T71rtmkqrkuz4ig8&topic_showPostId=t73pbk78p7dmzf3t#flow-post-t73pbk78p7dmzf3t)

JeroenDeDauw:
Thanks!

Needs a manual verification before merge.

I thought Maps had an API key for Google Maps... I'm wondering if it was removed because they changed their service in such a way that it's no longer needed. If so, they really should make up their mind already!

kghbln:
@JeroenDeDauw 

> Needs a manual verification before merge.

See https://github.com/JeroenDeDauw/Maps/pull/180#issuecomment-230304534

> I thought Maps had an API key for Google Maps... I'm wondering if it was removed because they changed their service in such a way that it's no longer needed. If so, they really should make up their mind already!

Yeah I remember, we had this once since Google required it for using the Maps from the very first hit to the API. Then they changed policy and now we are back to were we were before.

This reminds me that this should also be documented in the README or the document that holds such information right now.

Henryfunk:
Tested OK, thanks!

Should it perhaps be possible to add a version number? Google at any given time has three versions of Google Maps API for JS: a frozen version, a standard version and an experimental one. According to https://developers.google.com/maps/documentation/javascript/versions#version-rollover-and-version-types : "If you do not explicitly specify a version, you will receive the experimental version by default if you are on the standard plan (this includes customers who do not provide a key). If you're on the premium plan and don't specify a version, you will receive the release version by default."

JeroenDeDauw:
Thanks for testing. The version number thing sounds like a good idea. Can you create a new issue for it?

includes/services/GoogleMaps3/Maps_GoogleMaps3.php
```diff
@@ -318,6 +318,7 @@ protected function getDependencies() {
 	public static function getApiScript( $langCode, array $urlArgs = [] ) {
 		$urlArgs = array_merge(
 			[
+				'key' => $GLOBALS['egMapsGMaps3ApiKey'],
```
JeroenDeDauw: What will happen for existing users for which the maps work, but that do not have a key set? Won't this add `&key=`, which might then cause the map to fail?

Henryfunk: Yes, the key should only be added if a value for it is supplied.

PeterTheOne: fixed in 9b51cdb

includes/geocoders/Maps_GoogleGeocoder.php
```diff
@@ -35,7 +35,15 @@ public static function register() {
 	 * @return string
 	 */	
 	protected function getRequestUrl( $address ) {
-		return 'http://maps.googleapis.com/maps/api/geocode/xml?address=' . urlencode( $address ) . '&sensor=false';
+		$urlArgs = [
+			'address' => urlencode( $address ),
+			'sensor' => false
+		];
+		if ( !empty( $GLOBALS['egMapsGMaps3ApiKey'] ) ) {
```
JeroenDeDauw: I'll merge, but beware this it's better to do `$var !== ''`. The behaviour of `empty` is weird, and it makes the code less clear (as you can't immediately see the type of $var)

JeroenDeDauw:
Release notes need update

PR #190, opened 2016-08-01
add nominatim geocoder

### LINKED ISSUE #185: Geocoding with Nominatim
Hello,

It would be nice to be able to use [Nominatim](http://wiki.openstreetmap.org/wiki/Nominatim) for geocoding. In my city, it is much more accurate than Google Maps.

PeterTheOne:
fix #185

### DIFF
--- a/includes/geocoders/Maps_NominatimGeocoder.php
+++ b/includes/geocoders/Maps_NominatimGeocoder.php
@@ -5,6 +5,8 @@
  * 
  * Webservice documentation: http://wiki.openstreetmap.org/wiki/Nominatim
  *
+ * @since 4.0
+ *
  * @licence GNU GPL v2+
  * @author Peter Grassberger < petertheone@gmail.com >
  */
--- a/includes/geocoders/Maps_NominatimGeocoder.php
+++ b/includes/geocoders/Maps_NominatimGeocoder.php
@@ -39,7 +39,7 @@ final class MapsNominatimGeocoder extends \Maps\Geocoder {
 			'limit' => 1,
 		];
 
-		return 'http://nominatim.openstreetmap.org/search?' . wfArrayToCgi($urlArgs);
+		return 'https://nominatim.openstreetmap.org/search?' . wfArrayToCgi($urlArgs);
 	}
 	
 	/**
--- a/includes/geocoders/Maps_NominatimGeocoder.php
+++ b/includes/geocoders/Maps_NominatimGeocoder.php
@@ -1,0 +1,75 @@
+<?php
+
+/**
+ * Class for geocoding requests with the OSM Nominatim.
+ * 
+ * Webservice documentation: http://wiki.openstreetmap.org/wiki/Nominatim
+ *
+ * @licence GNU GPL v2+
+ * @author Peter Grassberger < petertheone@gmail.com >
+ */
+final class MapsNominatimGeocoder extends \Maps\Geocoder {
+	
+	/**
+	 * Registers the geocoder.
+	 * 
+	 * No LSB in pre-5.3 PHP *sigh*.
+	 * This is to be refactored as soon as php >=5.3 becomes acceptable.
+	 * 
+	 * @since 0.7
+	 */
+	public static function register() {
+		\Maps\Geocoders::registerGeocoder( 'nominatim', __CLASS__ );
+		return true;
+	}
+	
+	/**
+	 * @see \Maps\Geocoder::getRequestUrl
+	 * 
+	 * @since 0.7
+	 * 
+	 * @param string $address
+	 * 
+	 * @return string
+	 */	
+	protected function getRequestUrl( $address ) {
+		$urlArgs = [
+			'q' => urlencode( $address ),
+			'format' => 'jsonv2',
+			'limit' => 1,
+		];
+
+		return 'http://nominatim.openstreetmap.org/search?' . wfArrayToCgi($urlArgs);
+	}
+	
+	/**
+	 * @see \Maps\Geocoder::parseResponse
+	 * 
+	 * @since 0.7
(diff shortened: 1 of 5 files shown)


includes/geocoders/Maps_NominatimGeocoder.php
```diff
@@ -0,0 +1,75 @@
...
+	 * @return string
+	 */	
+	protected function getRequestUrl( $address ) {
+		$urlArgs = [
+			'q' => urlencode( $address ),
+			'format' => 'jsonv2',
+			'limit' => 1,
+		];
+
+		return 'http://nominatim.openstreetmap.org/search?' . wfArrayToCgi($urlArgs);
```
JeroenDeDauw: No HTTPS?

JeroenDeDauw:
Cool!

This is something you can easily write a test for. ... Or perhaps not entirely as `\Maps\Geocoder` is a pile of ****. Still, you could create a sane geocoding class and simply have an adapter deriving from `\Maps\Geocoder` that uses that.

```php
class Maps\Geocoders\NominatimGeocoder {
    function __construct( SomeWebRequestThing $internets ) { $this->internets = $internets; }
    function geocode( string $address ): LatLongValue {
       // weee
       $this->internets->request( /* ... */ );
       // weee
    }
}
```

Perhaps implementing some new `Maps\Geocoders\NonSuckyGeocoder` interface.

You can then have a unit test that verifies the URLs get build correctly given the input, and one that verifies various kinds of API responses get turned into the function return value. And to really test all the things, you could add a system test that runs some requests against the real API (this test would not be run by default, only with specific flag).

PeterTheOne:
I didn't feel like rewriting `\Maps\Geocoder` and I'm not really sure how to do better.

JeroenDeDauw:
My suggestion here does not involve rewriting anything already there. It describes how you can add the new functionality in such a way that it does not add to the pile of legacy crud already there.

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Maps\Test;
+
+/**
+ * @covers MapsNominatimGeocoder
+ *
+ * @since 4.0
```
JeroenDeDauw: These make no sense for non-package-public classes IMO

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @group Maps
+ *
+ * @licence GNU GPL v2+
+ * @author Peter Grassberger < petertheone@gmail.com >
+ */
+class NominatimGeocoderTest extends \PHPUnit_Framework_TestCase {
+
+	protected static function getMethod( $name ) {
+		$class = new \ReflectionClass( 'MapsNominatimGeocoder' );
```
JeroenDeDauw: Such reflection!

`MapsNominatimGeocoder::class`

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
...
+		];
+	}
+
+	/**
+	 * @covers MapsNominatimGeocoder::getRequestUrl
+	 *
+	 * @dataProvider addressProvider
+	 */
+	public function testGetRequestUrl( $address, $expected ) {
+		$getRequestUrl = self::getMethod('getRequestUrl');
```
JeroenDeDauw: spacing

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
...
+		$geocoder = new \MapsNominatimGeocoder( 'nominatim' );
+		$actual = $getRequestUrl->invokeArgs($geocoder, array( $address ));
```
JeroenDeDauw: spacing

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
...
+		];
+	}
+
+	/**
+	 * @covers MapsNominatimGeocoder::parseResponse
+	 *
+	 * @dataProvider responseProvider
+	 */
+	public function testParseResponse( $response, $expected ) {
+		$parseResponse = self::getMethod('parseResponse');
```
JeroenDeDauw: spacing

tests/phpunit/geocoders/NominatimGeocoderTest.php
```diff
@@ -0,0 +1,62 @@
...
+		$geocoder = new \MapsNominatimGeocoder( 'nominatim' );
+		$actual = $parseResponse->invokeArgs($geocoder, array( $response ));
```
JeroenDeDauw: spacing

JeroenDeDauw:
If you do not understand the approach I'm suggesting and think it would be instructive if I did it then I can give that a go.

PeterTheOne:
> If you do not understand the approach I'm suggesting and think it would be instructive if I did it then I can give that a go.

I think I know where you are going but I don't get why you would want to create an extra class and interface to wrap one new geocoder so we can test it if you say the Geocoder class is the problem. Why not restructure that one? In turn maybe I don't get it. :P If you want to do it, please do, I also think it could be instructive.

JeroenDeDauw:
Thanks. I'll submit a follow up

JeroenDeDauw:
Will have to be for after my holiday tho...

JeroenDeDauw:
Follow up at https://github.com/JeroenDeDauw/Maps/pull/214

PR #191, opened 2016-08-01
OpenLayers: Remove duplicate resizable

PeterTheOne:
See: https://github.com/JeroenDeDauw/Maps/blob/master/includes/services/OpenLayers/jquery.openlayers.js#L390 and https://github.com/JeroenDeDauw/Maps/blob/master/includes/services/OpenLayers/jquery.openlayers.js#L596


JeroenDeDauw:
Such gigantic function is gigantic

PR #192, opened 2016-08-01
add egMapsGMaps3ApiVersion number

### LINKED ISSUE #181: Option to add version number
It wiould be a good idea to add the option for users to supply a version number. Google at any given time has three versions of Google Maps API for JS: a frozen version, a standard version and an experimental one. According to https://developers.google.com/maps/documentation/javascript/versions#version-rollover-and-version-types : "If you do not explicitly specify a version, you will receive the experimental version by default if you are on the standard plan (this includes customers who do not provide a key). If you're on the premium plan and don't specify a version, you will receive the release version by default."

PeterTheOne:
fix #181

Haven't tested this yet.

### DIFF
--- a/Maps_Settings.php
+++ b/Maps_Settings.php
@@ -171,6 +171,9 @@ if ( !defined( 'MEDIAWIKI' ) ) {
 		// String. Google Maps v3 API Key
 		$GLOBALS['egMapsGMaps3ApiKey'] = '';
 
+		// String. Google Maps v3 API version number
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
Looks good. I'm fine with merging this if you make sure this gets tested before the next rel

PeterTheOne:
Ok, I tested it with `3.exp`, `3` and `3.24`, it loads different javascript files and map looks good.

JeroenDeDauw:
Thanks

PR #194, opened 2016-08-02
add leaflet.cluster

PeterTheOne:
Fixes: #141

Seems to work fine. Cluster icons are still missing, should we use the same ones as googlemaps3?

`spiderfy` and other options are not used for now.

TODO before merge:

- [x] add more options, including `spiderfy`
- [x] fix i18n
- [x] add icon
- [x] remove dontAddMarker
- [x] icon: show number of markers per cluster
- [x] fix conflict
- [x] test removing, recreating or refreshing cluster


includes/services/Leaflet/jquery.leaflet.js
```diff
@@ -25,7 +25,9 @@
 		* @param {Object} markerData Contains the fields lat, lon, title, text and icon
 		* @return {L.Marker}
 		*/
-		this.addMarker = function (properties) {
+		this.addMarker = function (properties, dontAddToMap) {
```
JeroenDeDauw: `dontAddToMap`? what? :) Is this some debug thing?

PeterTheOne: See this https://github.com/JeroenDeDauw/Maps/pull/194/files#diff-4a1096507c3fea5c0fb17f1231c4a617R179 vs. this
https://github.com/JeroenDeDauw/Maps/pull/194/files#diff-4a1096507c3fea5c0fb17f1231c4a617R184 .

For the cluster the markers are not directly added to the map but to the cluster thing. I could also rename it to `addToMap` if you want. ;)

PeterTheOne:
@kghbln I renamed `maps-googlemaps3-par-markercluster` in `en.json` and `qqq.json`, should I have renamed them in all the other languages also?

PeterTheOne:
Options comparison gm3 vs. leaflet:

- maps-googlemaps3-par-clustergridsize (default: 60) -> maxClusterRadius (default: 80)?
- maps-googlemaps3-par-clustermaxzoom -> disableClusteringAtZoom
- maps-googlemaps3-par-clusterzoomonclick -> zoomToBoundsOnClick?
- maps-googlemaps3-par-clusteraveragecenter -> n/a
- maps-googlemaps3-par-clusterminsize -> n/a
- n/a -> spiderfyOnMaxZoom (enabled by default)

kghbln:
> I renamed maps-googlemaps3-par-markercluster in en.json and qqq.json, should I have renamed them in all the other languages also?

I would do so since these messages do not need to be translated again in this case. Depending on the language it can take quite some time until a translation is available since there are sometimes only occasional translators around.

includes/services/Leaflet/jquery.leaflet.js
JeroenDeDauw: Right...

* addToMap would be nicer, esp than `!dontAddToMap`
* functions ought to do a single thing (on a given level of abstraction). If they take a boolean flag, you know they fail at this

PeterTheOne: fixed in bc6b2ef

PeterTheOne:
ready for merge

JeroenDeDauw:
Awesome \o/

PR #196, opened 2016-08-02
[DNM] add tilesurl option for leaflet map

### LINKED ISSUE #92: Allow alternate tile URLs for OSM tiles
The OSM tile URL is hard-coded to the primary OSM tile server over http (see [jquery.leaflet.js line 134](https://github.com/JeroenDeDauw/Maps/blob/master/includes/services/Leaflet/jquery.leaflet.js#L134)). It'd be nice if this could be changed via a setting (and maybe default to an https tile URL such as `https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png`

PeterTheOne:
fix #92

I don't have access to an alternative tile source, but as far as I know this should work.

### DIFF
--- a/Maps_Settings.php
+++ b/Maps_Settings.php
@@ -297,5 +297,9 @@ if ( !defined( 'MEDIAWIKI' ) ) {
 		// user does not provide one.
 		$GLOBALS['egMapsLeafletZoom'] = 14;
 
+		// String. The default tilesurl for the leaflet map. This value will only be used when the
+		// user does not provide one.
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
I'm somewhat concerned about security implications. Like this people can have images that they control be pulled in. And record what users are doing with the maps based on the requests.

Perhaps such things are already possible using other features of Maps? I don't recall.

Having this be disabled by default and requiring the site owners to opt-in would certainly ease my mind.

PeterTheOne:
Ah, haven't thought about that.

PeterTheOne:
I will change this and make it work more like the layers in the openlayers service.

JeroenDeDauw:
Collected some conflicts

PR #193, opened 2016-08-02
add leaflet.fullscreen

PeterTheOne:
fix #187

PR #201, opened 2016-08-03
rename enable-fullscreen option in all languages

PeterTheOne:
see also: #193


PeterTheOne:
6fcdcaf81a01a4fdbd797d25e2fbed8d3b5c2233 by @Nikerabbit already seems to fix this, thnx.

PR #211, opened 2016-08-11
Remove composer autoloading

Nikerabbit:
This breaks running composer test (and any other command) for MediaWiki core, because it tries to autoload Maps.php which then exits because `MEDIAWIKI` constant is not defined.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,9 +36,6 @@
 		"data-values/common": "~0.3.1"
 	},
 	"autoload": {
-		"files" : [
-			"Maps.php"
-		],
 		"psr-4": {
 			"Maps\\": "src/Maps/"
 		},


mwjames:
> This breaks running composer test (and any other command) for MediaWiki core, because it tries to autoload Maps.php which then exits because MEDIAWIKI constant is not defined.

How so? Where?

Luckily, Travis sets [0] so it doesn't become obvious but dependencies like SM would most likely suffer because suddenly Maps as required resource is no longer available and autoloaded.

[0] https://github.com/JeroenDeDauw/Maps/blob/master/build/travis/before_script.sh#L36

`echo 'require_once( __DIR__ . "/extensions/Maps/Maps.php" );' >> LocalSettings.php`

Nikerabbit:
It should be possible to reproduce this with steps like these (did not try myself):
1) Install MediaWiki
2) composer require mediawiki/semantic-maps
3) composer test

Loading Maps manually is not an issue for me. Not being able to test my patches before submitting while Maps (and probably other related extensions) are installed is.

mwjames:
PS: https://github.com/SemanticMediaWiki/WhatsNearby also requries Maps to be loaded and doesn't make any checks as it expects Composer to handle dependencies and manage the autoloading.

Nikerabbit:
You can consider this more of an issue report then. Clearly more work is needed for a full solution. Possible solutions I can think are
* stop autoloading extensions and favor the recommended extension registration approach
* make the extension start-up files not die when MediaWiki is not around

mwjames:
> stop autoloading extensions and favor the recommended extension registration approach

"registration approach" provides no means to load dependencies and you left with a hack approach [0, 1] that can blow up in our faces anytime (because some WMF developer decides to rock the boat and change the behaviour of used methods). Which brings me back to https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/1732#issuecomment-231953404.

> make the extension start-up files not die when MediaWiki is not around

Cannot say nothing about this, never looked at this closely enough.

[0] https://github.com/SemanticMediaWiki/SemanticGlossary/blob/master/SemanticGlossary.php#L22-L30
[1] https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/1367#issuecomment-195093756

JeroenDeDauw:
*sigh*... Thanks for reporting this issue @Nikerabbit. Personally I'm very reluctant to deal with this for all my extensions. Yet another MediaWiki change that requires a pile of changes and makes retaining compatibility very hard. yay

JeroenDeDauw:
I can see how this is a pain test wise... https://github.com/JeroenDeDauw/Maps/pull/230

kghbln:
I would like to avoid the autoloading too, however for another reason.

JeroenDeDauw:
@Nikerabbit is this still a problem for you after https://github.com/JeroenDeDauw/Maps/pull/260 ?

Nikerabbit:
I don't know really, I have Maps 3.8.2 which presumable does not include this fix and I get the following

```
composer test
> composer lint
> parallel-lint --exclude vendor
[Tue Oct 11 10:26:14 2016] [hphp] [25258:7f9700922400:0:000001] [] 
Fatal error: unknown class Job in /www/dev.translatewiki.net/docroot/w/extensions/SemanticMediaWiki/src/MediaWiki/Jobs/JobBase.php on line 18
Script parallel-lint --exclude vendor handling the lint event returned with an error
```

JeroenDeDauw:
No, that version indeed not have this change.

kghbln:
Is this still an issue?

JeroenDeDauw:
@kghbln that was not really the issue here

I'm going to close this since #260 probably fixed the issue. Please reopen if still present in 4.x.

PR #212, opened 2016-08-16
Update RELEASE-NOTES.md

JeroenDeDauw:
TODO:

* `Added Leaflet fullscreen control`: add small inline example (foo=bar)
* `Added OSM Nominatim Geocoder`: add small inline example (foo=bar)
* Google Maps API stuff should be mentioned in relevant install docs


JeroenDeDauw:
Looks like we don't have a real docs section on config right now. Then again, no config was really required so far. This seems like a good place to add a note that something needs to be done when one wants to use Google Maps: https://github.com/JeroenDeDauw/Maps/blob/master/INSTALL.md#configuration

PR #214, opened 2016-08-17
Refactor new geocoder and its test

JeroenDeDauw:
This includes adding of support of usage of the new interface and
proper dependency injection in the geocoding registration code.


JeroenDeDauw:
Presumably going to fail as I was not able to run the tests locally. Still some tests (and production code) missing for non-happy path stuff. Will add those when I got feedback from the tests.

I'm rather irked at not being able to run the tests locally without having MW installed as this code should not be depending on it. However the autoloading mechanism includes the Maps entry point, which does depend on MW. Not sure how to best resolve that.

JeroenDeDauw:
Forgot to cast to floats... that was simple to fix :)

PR #215, opened 2016-08-19
Add Leaflet layers

### LINKED ISSUE #2: Feature request: built-in OpenSeaMap support
A built-in support for the OpenSeaMap overlay [1] would be very nice. 

[1] http://wiki.openstreetmap.org/wiki/OpenLayers

PeterTheOne:
Alternative to #196.

Fixes #2 for Leaflet.

Uses the https://github.com/leaflet-extras/leaflet-providers Leaflet Plugin to load more alternative layers. Additionally adds MapQuest Layer.

This PR is still somewhat rough, feel free to provide feedback.

Known problems:

- [x] Not sure if it works with SemanticMaps yet, see: https://github.com/SemanticMediaWiki/SemanticMaps/pull/107
- [x] Doesn't handle Mapbox or HERE API keys
- [x] Missing i18n
- [x] Test if openlayers layers still work with these changes.
- [x] Change settings to boolean values array

/cc @krabina


PeterTheOne:
@JeroenDeDauw Can you take a look? Thanks :)

Maps_Settings.php
```diff
@@ -301,5 +301,60 @@
...
+			'Esri.WorldTopoMap',
+			'Esri.WorldImagery',
+			'Esri.WorldTerrain',
+			'Esri.WorldShadedRelief',
+			'Esri.WorldPhysical',
+			'Esri.OceanBasemap',
+			'Esri.NatGeoWorldMap',
+			'Esri.WorldGrayCanvas',
+			'MapQuestOpen',
+		];
```
JeroenDeDauw: It has occurred to me lately that while this is a sensible data structure for inside of the code, that for configuration it causes some issue: unless you're adding something, you'll need to redefine the array, or use some kind of filtering PHP code. So if the user wants to simply disable a single layer, the easiest thing for them to do is to copy the thing and remove the entry, which is kinda silly, and then also does not keep them up to date with the latest entries in new versions.

How about doing

```php
[
    'OpenStreetMap.BlackAndWhite' => true,
    'OpenStreetMap.HOT' => true,
]
```

Then they can easily disable one. You'd probably want to map it to a normal list when reading the config then, though that is really simple...

Thoughts @kghbln?

PeterTheOne:
Also there may be a bug when working with SemanticMaps, will look into it.

JeroenDeDauw:
> Also there may be a bug when working with SemanticMaps, will look into it.

What's the general state of this PR then? Is the "This PR is still somewhat rough" in the description current? I still need to actually look at it; the above comment was quite random.

> Indeed, I personally do not fancy "bulk" arrays either. 

Are you saying you rather not have 

```php
[
    'OpenStreetMap.BlackAndWhite' => true,
    'OpenStreetMap.HOT' => true,
]
```

either, and prefer having a setting for each layer?

PeterTheOne:
@JeroenDeDauw What is still rough and what is not working with SemanticMaps is the `evilOpenLayersHack` I tried to replicate.

kghbln:
> either, and prefer having a setting for each layer?

Generally I prefer having a setting each. It is perhaps a matter of how much extra programming this will require. Also having a setting each will probably be better in context of a configuration interface which might be invented at some time in the future.

pros:
- fine grained settings to manipulate behaviour
- individual documentation (one thing at a time)
- ready for configuration interfaces

cons:
- individual documentation (someone has to document this stuff)
- heaps of possible configuration work

In case it is possible to change things by just adding than this is also acceptable:

```php
[
    'OpenStreetMap.BlackAndWhite' => true,
    'OpenStreetMap.HOT' => true,
]
```

JeroenDeDauw:
Right. I concur with that reasoning. For the available mapping services for instance, having different settings would have been better. However here there really are a lot of layers, so the array with boolean values seems nicer.

PeterTheOne:
Ok, I will change this to a boolean values array.

includes/Maps_DisplayMapRenderer.php
```diff
@@ -145,6 +145,13 @@ protected function handleMarkerData( array &$params, Parser $parser ) {
 
 		if ( $params['mappingservice'] === 'openlayers' ) {
 			$params['layers'] = self::evilOpenLayersHack( $params['layers'] );
+		} else if ( $params['mappingservice'] === 'leaflet' ) {
+			global $egMapsLeafletLayerDependencies;
+			$layerDependencies = [];
+			$layerDependencies[] = "<script src='https://open.mapquestapi.com/sdk/leaflet/v2.2/mq-map.js?key=" . $GLOBALS['egMapsLeafletLayersApiKeys']['MapQuestOpen'] . "'></script>";
+			MapsMappingServices::getServiceInstance( 'leaflet' )->addLayerDependencies(
+				array_unique( $layerDependencies )
+			);
```
JeroenDeDauw: Why are you putting this code here? It seems to have little to do with the method and not need anything from it

JeroenDeDauw:
Looked at the code now, found one issue. Not as much code to review as I thought :)

PeterTheOne:
I have to test if openlayers layers still work with these changes.

includes/Maps_DisplayMapRenderer.php
```diff
@@ -225,28 +229,35 @@ public static function evilOpenLayersHack( $layers ) {
...
 		return $layerDefs;
 	}
 
-	/**
-	 * FIXME
-	 * @see evilOpenLayersHack
-	 */
-	private static function getLayerDependencies( array $layerNames ) {
-		global $egMapsOLLayerDependencies, $egMapsOLAvailableLayers;
+	public static function getLayerDependencies( $service, $params ) {
```
JeroenDeDauw: If this class is the only user of this method, it makes no sense for this to be public and static. Which makes me think it's probably used elsewhere, so the added param might break something.

includes/Maps_MappingService.php
```diff
@@ -248,6 +248,17 @@ public function addResourceModules( $modules ) {
 	}
 
 	/**
+	 * Adds the layer dependencies.
+	 *
+	 * @param array $dependencies
+	 */
+	public function addLayerDependencies( array $dependencies ) {
```
JeroenDeDauw: Why does this have Layer in its name? The code is not specific to layers at all. Or is it reasonable to expect it will be in the future?

includes/Maps_DisplayMapRenderer.php
PeterTheOne: Yes its used in `evilOpenLayersHack()` with was used in Semantic Maps. I did a change in https://github.com/SemanticMediaWiki/SemanticMaps/pull/107. So this change destroys some backward compatibility. Any Ideas how to prevent this?

JeroenDeDauw: I don't see any issues with SM requiring the latest Maps. Why would one want to run an older version anyway? For feature releases at least this makes little sense.

It's really weird that SM uses code from the class responsible for rendering of the displaymap parser function. Presumably something should be factored out.

If we merge the thing in, all of this gets easier to deal with.

includes/Maps_MappingService.php
PeterTheOne: I could name it `addDependencyArray` because `addDependencies( &$parserOrOut )` already exists and does something different.

includes/Maps_DisplayMapRenderer.php
PeterTheOne: `MapsLeaflet` deals with dependencies maybe code could be moved there..

includes/Maps_MappingService.php
JeroenDeDauw: > already exists and does something different.

Of course it does...

facepalm

`addHtmlDependency` and `addHtmlDependencies` seem like reasonable names for these. The existing `addDependencies` could be split into `addDependenciesToParser` and `addDependenciesToOutput`

PeterTheOne:
@JeroenDeDauw Can we merge this, or is it blocked by #235 ?

krabina:
https://github.com/JeroenDeDauw/Maps/issues/385

PR #225, opened 2016-09-06
prevent cluster creation when option not set

PeterTheOne:


JeroenDeDauw:
Follow up: https://github.com/JeroenDeDauw/Maps/pull/226

PeterTheOne:
A release makes sense. The bug was only being triggered external for example by the SemanticMaps ajax feature.

JeroenDeDauw:
Oh, then I'd not have bothered... since that stuff in SM is not released anyway.

And this means the release notes as misleading.... meh

Issue #229, opened 2016-09-13
Make unit tests runnable without having MediaWiki loaded


JeroenDeDauw:
Problem: http://stackoverflow.com/questions/39470775/using-composers-autoloader-without-including-files

PR #230, opened 2016-09-13
Allow running the unit tests without MediaWiki

JeroenDeDauw:
Fixes https://github.com/JeroenDeDauw/Maps/issues/229


JeroenDeDauw:
@mwjames would be awesome if we could also run the SMW unit tests without having MW installed. How about doing something similar? I checked the SMW test bootstrap, and things are not as simple there...

mwjames:
> @mwjames would be awesome if we could also run the SMW unit tests without having MW installed. How about doing something similar? I checked the SMW test bootstrap, and things are not as simple there...

I'm wondering what do we achieve by running it without MW? Are we concerned about speed of the tests? 

Writing unit tests is a nice practice because it provides you with some sort of confidence that a class works within the boundaries of its definition but I've seen that you can easily cheat your way through those tests without noticing that something broke along the way therefore I'm mostly concerned with integration tests (this is where you can actually observe that something works or is broken given the provided use cases) and for this to happen we need MW installed.

The test framework isn't broken and without some convincing arguments, I cannot find myself to invest time in changing the status quo.

JeroenDeDauw:
This allows people to clone the code, run composer install and be able to run the tests. No need to install a webserver, a database, MediaWiki and the relevant extensions.

Note that the difference between tests here is just those depending on MW and those that do not. You can have integration tests that do not depend on it, and you should, at least if the code does not suck by binding to MW (or whatever other framework) all over the place.

PR #234, opened 2016-09-14
fix openlayers centre bug

PeterTheOne:
fix https://github.com/SemanticMediaWiki/SemanticMaps/issues/80

I lost one line here: https://github.com/JeroenDeDauw/Maps/commit/a973fd93733983d7df24be8aac2f1633c20f8ff3#diff-5eb28ae85199d14351aaa7df78f774e6L96

### DIFF
--- a/includes/services/OpenLayers/jquery.openlayers.js
+++ b/includes/services/OpenLayers/jquery.openlayers.js
@@ -293,7 +293,7 @@
 		this.text( '' );
 
 		/**
-		 * ToDo: That layers being created by 'eval' will deny us the possiblity to
+		 * ToDo: That layers being created by 'eval' will deny us the possibility to
 		 * set certain options. It's possible to set properties of course but they will
 		 * show no effect since they are not passed as options to the constructor.
 		 * With this working we could adjust max/minScale to display overlays independent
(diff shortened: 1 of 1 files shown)


includes/services/OpenLayers/jquery.openlayers.js
```diff
@@ -100,6 +100,7 @@
 				markerData.lonlat.transform(new OpenLayers.Projection("EPSG:4326"), new OpenLayers.Projection("EPSG:900913"));
 			}
 
+			if (this.bounds != null) this.bounds.extend(markerData.lonlat); // Extend the bounds when no center is set.
```
JeroenDeDauw: The way I'm reading this code is that the bounds will get extended every time a marker is added. Is that incorrect?

PeterTheOne: That's how I understand it too and it's correct.

JeroenDeDauw: Does this mean that once the map has been fully loaded, and some event is causing addMarker to be called, the map will re-zoom and re-center?

(I don't really recall how the bounds stuff works, and if this rezoom and re-center happens if something gets added to it or if you need to make another call somewhere)

PeterTheOne: I guess. I tried to provoke something like this with the SemanticMaps ajax feature (which only loads and adds points that are within current bounds), but it worked fine.

JeroenDeDauw: Ok, going to trust you used the right incantations then

PR #260, opened 2016-10-09
Reordered initialization code

JeroenDeDauw:
This adds settings to disable Maps and to disable its SMW integration.
It also removes everything but global declarations from code immediately
executed, so should allow for inclusion of Maps via Composer even without
MediaWiki being loaded.

### DIFF
--- a/Maps.php
+++ b/Maps.php
@@ -1,4 +1,5 @@
 <?php
+
 /**
  * Initialization file for the Maps extension.
  *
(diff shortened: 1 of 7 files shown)


JeroenDeDauw:
@kghbln you got the setting you where asking for :)

JeroenDeDauw:
@mwjames this is pretty cool actually, as it allows inclusion of the extension via Composer without MW being loaded. So I can just clone the repo here, run composer install, and run phpunit and have it run all MW independent tests. I'd love that for SMW as well

mwjames:
> So I can just clone the repo here, run composer install, and run phpunit and have it run all MW independent tests. I'd love that for SMW as well

Expect from the stuff below, I think you already can do that but I'm not sure how helpful this is given that there are not many tests that do not rely on MW or MW reference objects (such as Title).

```
if ( !defined( 'MEDIAWIKI' ) ) {
	die( 'Not an entry point.' );
}
```

JeroenDeDauw:
Well, it's good to have more such tests (and production code). Framework binding is costly

kghbln:
> @kghbln you got the setting you where asking for :)

heureka

That's awesome since I still have some wikis running v2.x which is bad.

So will I have to set `$egMapsDisableExtension` to "true" or "false" to disable loading? I guess it is "true".

JeroenDeDauw:
Yeah to `true`, it's not disabled by default :)

kghbln:
Thanks for confirming!

PR #264, opened 2016-10-09
Moved i18n registration out of wgExtensionFunctions

JeroenDeDauw:


JeroenDeDauw:
Note: if this change is needed, then also change https://github.com/JeroenDeDauw/Validator/blob/master/Validator.php

kghbln:
> Note: if this change is needed, then also change ...

Done with https://github.com/JeroenDeDauw/Validator/pull/23

PR #266, opened 2016-10-11
Remove SF form input

yaronkoren:
SF now defines its own map form inputs


JeroenDeDauw:
Thanks

Missing some stuff:

Maps_Settings.php: $GLOBALS['egMapsDefaultServices']['fi'] = $GLOBALS['egMapsDefaultService'];

JeroenDeDauw:
While I'm personally fine with this change, I don't recall running this by the users. Did you already ask on the mailing list if people are fine with this? I'd like to provide a chance where objections can be voiced.

yaronkoren:
I didn't ask. I assume no one would care, since this is an internal code issue, not something that would affect the functionality (even if the behavior of the two Google Maps inputs is slightly different, that's something that can be fixed in the code if people prefer the Semantic Maps version). The general look-and-feel is the same.

yaronkoren:
Thanks for pointing out the remaining files; hopefully the current patch covers everything...

JeroenDeDauw:
Right, thinking about it, nothing would change for the users... Or at least not much. Anyway, no need to ask then. Is there anything users might need to pay attention though? Might be good to put that into the release notes.

yaronkoren:
I can't think of anything... currently, SF checks for the existence of SM_VERSION to see whether it should define its own maps inputs. If this patch is accepted, I'll change the SF code to also check whether the Maps version is < 4.0. Ideally, users won't notice the difference - that they're now using the SF version of the Google Maps input instead of the SM/Maps one.

JeroenDeDauw:
Can't you check if there already is a map input registered or something? Checking the Maps version seems a bit roundabout. I don't care much though, just a suggestion.

yaronkoren:
Thank you! I thought about checking for duplicate registration, but my guess was that it would be impossible, given that SF registers its own input types first. I could be wrong, though; I didn't look into it.

Issue #250, opened 2016-10-05
Make `$egGoogleJsApiKey` an alias setting to `$egMapsGMaps3ApiKey`


kghbln:
This will probably make it easier for upgraders from Maps 2.x since they will not have to change this setting. This is related to https://github.com/JeroenDeDauw/Maps/issues/179 which was released with Maps 3.8 I got this idea from this [thread](https://www.mediawiki.org/wiki/Topic:Tb9wyrsmyuxnnob4). Perhaps it is not too complicated to implement this. @JeroenDeDauw

JeroenDeDauw:
Ah, good idea

kghbln:
Probably, but in the meantime we also have an API version configuration parameter so to keep the naming consistent ...

PR #269, opened 2016-10-12
Added compat for old egGoogleJsApiKey setting

JeroenDeDauw:
Fixes https://github.com/JeroenDeDauw/Maps/issues/250

### DIFF
--- a/Maps.php
+++ b/Maps.php
@@ -166,6 +166,10 @@ $GLOBALS['wgExtensionFunctions'][] = function () {
 
 
 	// Google Maps API v3
+	if ( $GLOBALS['egMapsGMaps3ApiKey'] === '' && array_key_exists( 'egGoogleJsApiKey', $GLOBALS ) ) {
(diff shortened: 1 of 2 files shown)


JeroenDeDauw:
@kghbln are we actually dealing with the same API key here though? Or will people still need to update the value of the setting if they have it from Maps 2.x?

kghbln:
Well I am assuming here that Google does not require to change an existing key. This would be bull... imho so I did not even consider this as an potential issue. To tell the truth: I do not know how the keys looked like back then and how they look like now.

JeroenDeDauw:
Well... that's kinda relevant :)

kghbln:
Well it depends if the key was generated for v2 or v3. If it was generated for v2 one [needs to get a new key](https://developers.google.com/maps/documentation/javascript/v2tov3#newkey). However I believe that the v2 no longer works and people using a key via the precursor setting are already using a v3 key. So yes it should be the same key.

PR #301, opened 2017-01-09
Maps_GoogleGeocoder - address being double urlencoded

somescout:
Address is double urlencoded:
first time in $urlArgs initialization
second time in wfArrayToCgi function

That breaks it for Unicode-encoded languages (specifically Russian)


JeroenDeDauw:
Thanks!

PR #325, opened 2017-05-10
[DNM] Use latest version of Leaflet

JeroenDeDauw:
Not getting this to work properly...
Been trying for an hour to get the resources loaded, but my wiki
is not allowing access to files in vendor/ it seems. Might be that
it is configured badly. I don't want to spend more time on MW shit,
though would be awesome if someone figures out what the issue is.


JeroenDeDauw:
@mwjames you any idea how to make this work?

@PeterTheOne FYI

kghbln:
@JeroenDeDauw What is the difference to #326 ?

JeroenDeDauw:
With this change the leaflet code is not included in the Maps repo but pulled in via Composer. Just going to close this, someone can have another go if they want.

kghbln:
Ah, this would indeed have been nice. :|

PeterTheOne:
Great idea, but was the vendor path something that is reachable for the web browser?

JeroenDeDauw:
I suppose not, looking at what I wrote. I don't really recall though

PR #326, opened 2017-05-10
Upgrade Leaflet to 1.0.3

JeroenDeDauw:


JeroenDeDauw:
Second attempt after https://github.com/JeroenDeDauw/Maps/pull/325

Markers not showing properly. This is a single marker: failmakers

No idea what the issue is and quite sick of this shit now. Upgrading this should not take an hour. Yay MediaWiki development

PR #346, opened 2017-06-10
4.3.0 release

JeroenDeDauw:


Maps.php
```diff
@@ -70,7 +70,7 @@
 		);
 	}
 
-	define( 'Maps_VERSION' , '4.3 alpha' );
+	define( 'Maps_VERSION' , '4.3' );
```
kghbln: 4.3 is so marketing. ;) Why not 4.3.0 for consistency.

JeroenDeDauw: lazy Jeroen does not really care about the number here

kghbln: Bunny is sad now.
sadbunny

Issue #351, opened 2017-07-10
implement Leaflet static option


PeterTheOne:
Leaflet static option doesn't work, for googlemaps3 it looks like this: https://www.semantic-mediawiki.org/wiki/Maps_examples/Google_Maps_with_a_static_map

kghbln:
Implemented with 4.4.0 and already [documented.](https://www.semantic-mediawiki.org/wiki/Maps_examples/Leaflet_with_a_static_map)

PR #354, opened 2017-08-08
Fix leaflet marker custom icons (#304) and implement leaflet static map option (#351)

### LINKED ISSUE #304: Custom Leaflet markers are misaligned
### Setup

- MW version: 1.28.0
- DB (MySQL etc.): MySQL 5.6.30
- PHP version: 5.6.29-0+deb8u1
- Maps version: 4.0.4
- SMW version (if applicable): 2.4.5

### Issue

When using Leaflet with a custom marker icon with (with `icon` or `activeicon`), the top left corner is always used has the center so the result is quite ugly:
capture d ecran 2017-01-31 a 19 58 07

Leaflet has several icon properties that can be used to tweak this: http://leafletjs.com/examples/custom-icons/#creating-an-icon
But they currently can't be set from MediaWiki.

hidrarga:
Fix #304 and implement #351

### DIFF
--- a/includes/services/Leaflet/jquery.leaflet.js
+++ b/includes/services/Leaflet/jquery.leaflet.js
@@ -251,6 +251,10 @@
 					position: 'topleft'
 				};
 			}
+			
+			if (options.static) {
+				mapOptions.scrollWheelZoom = false;
+			}
 
 			var map = L.map( this.get(0), mapOptions ).fitWorld();
 			this.map = map;
--- a/includes/services/Leaflet/jquery.leaflet.js
+++ b/includes/services/Leaflet/jquery.leaflet.js
@@ -254,6 +254,13 @@
 			
 			if (options.static) {
 				mapOptions.scrollWheelZoom = false;
+				mapOptions.doubleClickZoom = false;
+				mapOptions.touchZoom = false;
+				mapOptions.boxZoom = false;
+				mapOptions.tap = false;
+				mapOptions.keyboard = false;
+				mapOptions.zoomControl = false;
+				mapOptions.dragging = false;
 			}
 
 			var map = L.map( this.get(0), mapOptions ).fitWorld();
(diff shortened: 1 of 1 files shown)


kghbln:
@hidrarga Thanks for the patch! I just requested reviews for it.

JeroenDeDauw:
The icon change looks good but would be cool if someone could verify it

The static map change I am not sure about. Is disabling scrollweel zoom sufficient to make a map "static"? I was under the impression there was a dedicated option for disabling scollwheel zoom. For whatever reason neither that option nor the static one are listed in the docs though https://www.semantic-mediawiki.org/wiki/Extension:Maps/Displaying_maps and my memory on this is quite fuzzy.

hidrarga:
You are right for the static map option. I did this a bit too fast, and now that I look closer at the examples for the Openlayers and Googlemaps services, the resulting map looks like a static image. I can disable more controls to have the same behaviour. 

So I would disable : 
* doubleClickZoom
* boxZoom
* touchZoom
* tap
* keyboard
* zoomControl
* dragging

However, I must say that I would find useful to keep the zoom controls enabled, so that the user can still zoom on the map.

JeroenDeDauw:
> However, I must say that I would find useful to keep the zoom controls enabled, so that the user can still zoom on the map.

If what you want behavior wise is what you already did in this PR then I'd like to see this done as an option that disable zooming with the scrollwheel.

> I can disable more controls to have the same behaviour.

Sounds good

hidrarga:
>If what you want behavior wise is what you already did in this PR then I'd like to see this done as an option that disable zooming with the scrollwheel.

Do you have any idea of how we could name this option ? I'm not aware of the naming convention regarding the map options... Also, I suppose this will be a common option to every map service ?

JeroenDeDauw:
I though there was a "scrollwheelzoom" option. There is no explicit convention but you can look at the existing options https://www.semantic-mediawiki.org/wiki/Extension:Maps/Displaying_maps

> Also, I suppose this will be a common option to every map service ?

If you want to add it to all of them. If not, it can be just for one service

JeroenDeDauw:
I've merged this now, updated the release notes https://github.com/JeroenDeDauw/Maps/blob/master/RELEASE-NOTES.md and added the missing static parameter to the documentation https://www.semantic-mediawiki.org/wiki/Extension:Maps/Displaying_maps#Base_parameters

The other changes are also welcome

PeterTheOne:
Nice PR, thnx! :)

Issue #324, opened 2017-05-10
Upgrade Geocoders to new interface


JeroenDeDauw:
New interface is at `src/Geocoders/Geocoder.php`. Example implementation is at `src/Geocoders/NomatimGeocoder.php` with test at `tests/Unit/Geocoders/NominatimGeocoderTest.php`.

Still need to be migrated:

* includes/geocoders/Maps_GeocoderusGeocoder.php
* includes/geocoders/Maps_GeonamesGeocoder.php (most important)
* includes/geocoders/Maps_GoogleGeocoder.php

JeroenDeDauw:
Done

Issue #294, opened 2016-12-06
Features to be removed in 5.0


JeroenDeDauw:
This issue is for collecting features that are not used or broken altogether that can thus be removed without affecting essentially any users. Those features take time away from the ones that actually matter, so it is better to drop them.

---

* ~~[Broken custom layers functionality](https://github.com/JeroenDeDauw/Maps/issues/176) - Removed in Maps 3.7~~
* [Broken geocoding service selection](https://github.com/JeroenDeDauw/Maps/issues/285) - Partially fixed, but presumably can just be removed
* GeocoderUs service seems to be down: http://geocoder.us/service/rest/?address=New+York

PR #358, opened 2017-09-22
Better geocoding cache using MediaWiki cache

hidrarga:
Hello,

I'm proposing a new approach for the geocoding cache system. It makes use of the MediaWiki cache. 

The issue I had with the current geocoding cache is that it was an in-memory cache. Thus, it wasn't persistent.

As I'm not an expert in MediaWiki development, I took my inspiration from the example provided by the [Memcache manual page](https://www.mediawiki.org/wiki/Memcached#Using_memcached_in_your_code). Note that one doesn't need to use memcache for this solution to work: I retrieve any cache in use, and I work with what I get. 

I store the coordinates with an expiry time of one day. When there is no coordinates to store (because geocoding failed), the result is not stored in cache [as explained in the documentation](https://doc.wikimedia.org/mediawiki-core/master/php/classBagOStuff.html#a79e635258ff6ee2167df75909067db5a).

Finally, I changed the location of the cache (method `geocode` of class `Geocoder` instead of class `Geocoders`) because I wanted to cache the result of line 194 of Geocoders.php:
```php
$coordinates = $geocoder->geocode( str_replace( ',', '', $address ) );
```
I would appreciate any review on my code :) As I said, I'm not an expert of the MediaWiki framework, so some mistakes are still possible.

PS: What should we do with the now unused cacheRead, cacheWrite, hasGlobalCacheSupport, etc. method ?


JeroenDeDauw:
Thanks a lot for working on this!

Unfortunately the code you touched is deprecated (not mentioned properly, sorry for that) and being refactored away in favor of the better designed code at https://github.com/JeroenDeDauw/Maps/tree/master/src/Geocoders

A clean way to do the caching would be to create a caching [decorator](https://en.wikipedia.org/wiki/Decorator_pattern) for the new `Geocoder` interface in the `src` folder. In this decorator you could then put your MediaWiki caching code, though you could also use the [MediaWikiCache](https://github.com/JeroenDeDauw/SimpleCache/blob/master/src/Cache/MediaWikiCache.php) implementation of `SimpleCache\Cache`, as Maps already uses the SimpleCache library.

If you are not used to composition and dependency injection what I wrote might be hard to understand. If so, let me know, and I will give it a go myself.

hidrarga:
Ah indeed, I didn't pay enough attention here... It's not a real problem for me, as I needed this caching system for my wiki anyway. But of course, I'm willing to implement this feature so that everybody can benefit from it. 

I understood how to implement the decorator, but I don't know how to call it in the code. In the `Geocoders` class, the `geocode` method is actually taking the old `Geocode` object in argument... So, I guess I would need to make my `GeocodeCache` (or whatever we call it) implement the new `Geocode` interface. Then, I would pass the object into the `MapsOldGeocoderAdapter` ? It's becoming quite ugly ^^' Any suggestions ? Is there something I'm missing ?

I'm not used to php dependency injection, but I understood the global idea.

JeroenDeDauw:
Will reply later

Related issue: https://github.com/JeroenDeDauw/Maps/issues/324

JeroenDeDauw:
First of all, beware that not all geocoders are in the new format yet. This ticket describes that more: https://github.com/JeroenDeDauw/Maps/issues/324

The geocoders are registered in Maps.php:

```php
	// Registration of the GeoNames service geocoder.
	$GLOBALS['wgHooks']['GeocoderFirstCallInit'][] = 'MapsGeonamesGeocoder::register';

	// Registration of the Google Geocoding (v2) service geocoder.
	$GLOBALS['wgHooks']['GeocoderFirstCallInit'][] = 'MapsGoogleGeocoder::register';

	// Registration of the geocoder.us service geocoder.
	$GLOBALS['wgHooks']['GeocoderFirstCallInit'][] = 'MapsGeocoderusGeocoder::register';

	// Registration of the OSM Nominatim service geocoder.
	$GLOBALS['wgHooks']['GeocoderFirstCallInit'][] = function() {
		\Maps\Geocoders::registerGeocoder(
			'nominatim',
			new \Maps\Geocoders\NominatimGeocoder( new SimpleFileFetcher() )
		);
		return true;
	};
```

The last one is registration of the geocoder that uses the new interface. It just calls `\Maps\Geocoders::registerGeocoder` with an implementation of the new interface, so no need to deal with `MapsOldGeocoderAdapter`. You could wrap the geocoder on this point, like this:

```php
	$GLOBALS['wgHooks']['GeocoderFirstCallInit'][] = function() {
		\Maps\Geocoders::registerGeocoder(
			'nominatim',
			new CachingGeocoderDecorator(
                             new \Maps\Geocoders\NominatimGeocoder( new SimpleFileFetcher() ),
                             new \SimpleCache\MediaWikiCache()
                        )
		);
		return true;
	};
```

It is quite unfortunate that at present this needs to be repeated for each geocoder unless you do some more refactoring. My idea is to turn `Geocoders` into a thin facade or (better, but perhaps too difficult) ditch it completely. The correct instance of the new interface would just be injected where it is needed. ... ok, the important bit here is: I think there is [no need for "supporting multiple geocoders at the same time"](https://github.com/JeroenDeDauw/Maps/issues/294). If that support gets dropped (which would make the described refactor easier), then the duplication problem goes away.

JeroenDeDauw:
Ugh. I suppose it is better to decorate the geocoder just before it is stuffed in the MapsOldGeocoderAdapter for now. In other words, changing Geocoders.php:

```php
$geocoder = new MapsOldGeocoderAdapter(
	self::$registeredGeocoders[$geocoderIdentifier],
	$geocoderIdentifier
);
```

into

```php
$geocoder = new MapsOldGeocoderAdapter(
	new CachingGeocoderDecorator(
            self::$registeredGeocoders[$geocoderIdentifier],
            new \SimpleCache\MediaWikiCache()
        ),
	$geocoderIdentifier
);
```

JeroenDeDauw:
I was bored and tired so refactored one of the remaining geocoders to use the new interface: https://github.com/JeroenDeDauw/Maps/pull/363

JeroenDeDauw:
Now completed the refactoring away of the old Geocoding class hierarchy, so dealing the with geocoding code is a bunch simpler now: https://github.com/JeroenDeDauw/Maps/pull/364

hidrarga:
Wow, many thanks for this ! Yes, this made the implementation of the caching system really easy :)

I had a look at the MediaWikiCache implementation that you mentioned. It was easier for me to keep using BagOStuff though, because I wanted to store the coordinates even when they equal `null`, i.e we cannot fetch the coordinates (because of too many requests, geocoding server is down, or the address is just wrong). There is no need to spam the server here, so imho it's better to wait till next day to retry eventually.

src/Geocoders/CachingGeocoder.php
```diff
@@ -0,0 +1,45 @@
...
+		$this->cache = $cache;
+	}
+	
+	/**
+	 * @param string $address
+	 *
+	 * @return LatLongValue|null
+	 */
+	public function geocode( $address ) {
+		$coordinates = $this->cache->get( $address );
```
JeroenDeDauw: I think some kind of namespacing is needed here. Like this you allow users to retrieve whatever cache item they know the keys off, which at best allows them to crash the software by retrieving something that is not a LatLongValue, and at worst allows them to access information they should not have access to.

src/Geocoders/CachingGeocoder.php
```diff
@@ -0,0 +1,45 @@
...
+	protected $geocoder;
+	
+	protected $cache;
+
+	/**
+	 * @param Geocoder $geocoder
+	 * @param BagOStuff $cache
+	 *
+	 */
+	public function __construct( $geocoder, $cache ) {
```
JeroenDeDauw: No docs are needed here, you can just have

```php
public function __construct( Geocoder $geocoder, BagOStuff $cache ) {
```

src/Geocoders/CachingGeocoder.php
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @since 5.0
+ *
+ * @licence GNU GPL v2+
+ * @author HgO < hgo@batato.be >
+ */
+class CachingGeocoder implements Geocoder {
+	protected $geocoder;
+	
+	protected $cache;
```
JeroenDeDauw: The MediaWiki style of having thins be protected by default, rather than private by default, is harmful in my opinion.

includes/Geocoders.php
```diff
@@ -153,45 +144,18 @@ public static function geocode( $address, $geoService = '' ) {
...
-
-		// Add the obtained coordinates to the cache when there is a result and the cache is enabled.
-		if ( $egMapsEnableGeoCache && $coordinates ) {
-			self::$globalGeocoderCache[$address] = $coordinates;
-		}
+		
+		$caching = new Geocoders\CachingGeocoder( 
+			$geocoder, 
+			wfGetCache( CACHE_ANYTHING )
+		);
```
JeroenDeDauw: Like this for every call to the geocode function, a new instance of the CachingGeocoder is created. Perhaps not a huge deal, but there is some irony here :)

JeroenDeDauw:
Great work! This can be merged as soon as the namespacing issue is resolved. My other comments are basically about code style and are not that important.

> I'm not used to contributing to open source projects, so let me know if I did something wrong there.

All good!

> I had a look at the MediaWikiCache implementation that you mentioned. It was easier for me to keep using BagOStuff though

Sure, that makes sense. I might swap this out in the future to make the geocoding code fully independent from MediaWiki, and to make writing automated tests for it easier.

src/Geocoders/CachingGeocoder.php
hidrarga: Oh yes, you are right ! It's should be fine now :)

includes/Geocoders.php
hidrarga: Aha indeed, that's funny ^^ But doing it in another way would complicate things, no ? I don't see a nice way actually, as the input `Geocoder` might change at each call...

hidrarga:
Thanks for your help ! :)

>Sure, that makes sense. I might swap this out in the future to make the geocoding code fully independent from MediaWiki, and to make writing automated tests for it easier.

That would be very nice indeed :)

JeroenDeDauw:
\o/ I've updated the release notes: https://github.com/JeroenDeDauw/Maps/blob/master/RELEASE-NOTES.md#maps-500

PR #373, opened 2017-10-04
Fatal error when calling geocode parser function

hidrarga:
I noticed a fatal error when we want to call the `geocode` parser function. You can see an example here : https://sandbox.semantic-mediawiki.org/wiki/Seattle


JeroenDeDauw:
Nice catch. Not sure why I did not run into this, as I did a manual test with the parser function. Just verified we cannot inject the service... seems like some more legacy crud needs to be killed before that is possible.

PR #375, opened 2017-10-10
Apply IDE CS to all the code

JeroenDeDauw:


SemanticMaps/SemanticMaps.hooks.php
```diff
@@ -82,7 +86,8 @@ public static function addGeoCoordsDefaultFormat( &$format, array $printRequests
 				$hasCoords = false;
 
 				// Loop through the print requests to determine their types.
-				foreach( $printRequests as /* SMWPrintRequest */ $printRequest ) {
+				foreach ( $printRequests as /* SMWPrintRequest */
+						  $printRequest ) {
```
Nikerabbit: Eww alignment

src/Semantic/DataValues/CoordinateValue.php
```diff
@@ -127,7 +76,8 @@ protected function parseUserValue( $value ) {
 
 		if ( $value === '' ) {
 			$this->addError( wfMessage( 'smw_novalues' )->text() );
-		} else {
+		}
+		else {
```
Nikerabbit: Any particular reason to diverge from the common MediaWiki style of not adding newline before else/elseif?

Maps.php
```diff
@@ -47,8 +47,7 @@
 $GLOBALS['wgExtensionMessagesFiles']['MapsMagic'] = __DIR__ . '/Maps.i18n.magic.php';
 $GLOBALS['wgExtensionMessagesFiles']['MapsAlias'] = __DIR__ . '/Maps.i18n.alias.php';
 
-
-$GLOBALS['wgExtensionFunctions'][] = function () {
+$GLOBALS['wgExtensionFunctions'][] = function() {
```
Nikerabbit: In common MediaWiki style there is space after anonymous functions, just like with `if` to indicate it's not a function call.

SemanticMaps/SemanticMaps.hooks.php
JeroenDeDauw: wtf... guess the IDE got confused by the weird inline comment. Thanks for pointing this out, I'll remove the comment so it does not get confused.

src/Semantic/DataValues/CoordinateValue.php
JeroenDeDauw: This form seemed more common in the code written by the WMDE FUN team. I used the style created for those projects (https://github.com/wmde/intellij-settings) here. Then again I might have picked the wrong option. Still hammering out the IDE rules.

MWs PHPCS does not complain about either form, and perhaps it does not matter that much.

Maps.php
JeroenDeDauw: Thanks. Will change this. Seems like we have a slight skew to the less spacy form in the FUN code, but the usage is inconsistent, and we apparently don't have many anon functions.

JeroenDeDauw: Actually this is weird in MW style then...

    function noSpaceHere() {

vs

    $whySpace = function () {

So in case of anon declarations you have a space and for named ones you do not.

Nikerabbit: Just a different point of view: `function` is a keyword and there is a space after keyword.

JeroenDeDauw: Right. Thanks for explaining

PR #377, opened 2017-10-11
Fixes various links to smw.o

kghbln:


tests/Integration/parsers/LocationParserTest.php
```diff
@@ -64,8 +64,7 @@ public function testGivenTitleThatIsLink_linkIsSetAndTitleIsNot( $link ) {
 
 	public function linkProvider() {
 		return [
-			[ 'https://semantic-mediawiki.org' ],
-			[ 'http://www.semantic-mediawiki.org' ],
```
JeroenDeDauw: Would have been better to keep this. It is test code. Could just have well had test.com instead of smw.org.

JeroenDeDauw:
Thanks!

tests/Integration/parsers/LocationParserTest.php
kghbln: I was indeed a bit wondering about this. Well two instead of three test links should do, too.

PR #380, opened 2017-10-18
Dynamically resize leaflet custom icons

### LINKED ISSUE #360: Custom markers are resized to 25x41 px
### Setup

- MW version: 1.29.1
- DB (MySQL etc.): 5.5.57
- PHP version: 5.6.31-1~dotdeb+7.1
- Maps version: 4.4.0
- SMW version (if applicable): 2.5.4

### Issue

I'm creating a new bug following https://github.com/JeroenDeDauw/Maps/issues/304#issuecomment-331645611 because this is not fixed.

We have custom markers that are smaller than the regular Leaflet markers:
oval-15

Since Maps 4.4.0, the marker size is forced to 25x41 px (Leaflet's default size) so our markers are resized to this size:
image

As a workaround, I created 25x41 px marker images with blank padding, it's better but now the tooltip is too high (and the hitbox is bigger than the visible marker, which is confusing for users):
image

hidrarga:
This fixes #360. 

Because the marker object loads the Leaflet's default icon when no icon is provided, and because we must wait until the custom icon is loaded to know its size, I'm setting the opacity of the marker to 0. Otherwise, the default icon might appear subliminally.

For the position of the popup, I found that 2/3 of the icon's height looked nice. I'm also shifting the popup by 1px to the left when the width is odd. Those choices are of course quite arbitrary, so I'm open for suggestions ;)

### DIFF
--- a/includes/services/Leaflet/jquery.leaflet.js
+++ b/includes/services/Leaflet/jquery.leaflet.js
@@ -28,23 +28,32 @@
 		this.createMarker = function (properties) {
 			this.points.push( new L.LatLng(properties.lat, properties.lon) );
 
-			if (!properties.hasOwnProperty('icon') || properties.icon === '') {
-				var icon = new L.Icon.Default();
-			} else {
-				var iconOptions = L.Icon.Default.prototype.options;
-				iconOptions.iconUrl = properties.icon;
-
-				var icon = new L.Icon(iconOptions);
-			}
-
 			var markerOptions = {
-				title:properties.title,
-				icon:icon
+				title:properties.title
 			};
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Awesome!

@Rudloff might want to try this out

Would be good if someone other than hidrarga can test this and report back.

Rudloff:
I tried the patch and it works correctly for my use case :)

JeroenDeDauw:
Released as 5.0.3

Rudloff:
This looks really nice on our wiki now, thanks!

PR #386, opened 2017-11-02
Leaflet attribution

### LINKED ISSUE #384: Map attribution not shown
### Setup

- Maps version: 5.0.2
- MW version: 1.27.3

### Issue
Maps show "Leaflet | © OpenStreetMap contributors" as attribution no matter what is set as custom attribution.

e.g. usually, custom attributions are set
 https://github.com/JeroenDeDauw/Maps/blob/9a35aad579d0e2f733bce4a62281390172ff8680/includes/services/Leaflet/leaflet-providers/leaflet-providers.js#L534

and the default should get replaced here:
https://github.com/JeroenDeDauw/Maps/blob/9a35aad579d0e2f733bce4a62281390172ff8680/includes/services/Leaflet/leaflet-providers/leaflet-providers.js#L71-L84

but anyway, this does not seem to work:
http://standards.kdz.eu/index.php?title=MapsTest

PeterTheOne:
fix #384 

The `leaflet-providers` plugin doesn't need our attribution string, so we don't pass it.
`OpenStreetMap` is also included in the `leaflet-providers` plugin, so we just use that, also helps with combined attribution strings with layers and overlaylayers. For this to work `leaflet-providers` needs to also be loaded for the default `OpenStreetMap`.
Also `MapQuestOpen` has its own attribution functionality so we don't need to pass an attribution string.


krabina:
seems to be a problem in Semantic Maps as well...

PeterTheOne:
This should also fix it for Semantic MediaWiki..

JeroenDeDauw:
@PeterTheOne @krabina Is either of you running the latest version of Maps with PHP 5.x? I'm asking because the latest development version now requires PHP 7.x and I'm wondering if there is reason to backport changes such as this one.

PeterTheOne:
My vagrant dev test wiki runs PHP 5.x, i will need to update it..

krabina:
I have a couple of wikis stilll on PHP 5...

JeroenDeDauw:
@PeterTheOne perhaps you can use https://github.com/addshore/mediawiki-docker-dev with some changes:

image

@krabina And for those do you need these latest Maps changes?

krabina:
not, not really. I am plannung to move to PHP 7 soon...

krabina:
Found another problem: when I click on "basemap.at" which is one of the available layers, the link points to:
 /WIKIROOT/www.basemap.at 
and not to http://www.basemap.at

PR #393, opened 2017-11-16
Leaflet multi layers

### LINKED ISSUE #387: Feature: Layer Groups and Layers Control for leaflet
leaflet Layer Groups and Layers Control should be supported: http://leafletjs.com/examples/layers-control/

PeterTheOne:
fixes #387 // cc @krabina 

Adds support for multiple layers with leaflet. Displays layer controls like [this leaflet example](http://leafletjs.com/examples/layers-control/).

Only displays `L.control.layers` when there is more than one baselayer, control doesn't show up if there is for example a baselayer and an overlaylayer. overlaylayer are nevertheless included in the control. Should the default behavior be different? I could display the control once there is at least another baselayer or an overlaylayer. And should there be an extra option, like `forceLayerControl` (makes sense for the current default), `hideLayerControl` (usefull for the other default I just described)?

@JeroenDeDauw Did I do the aliasing of `layer` / `layers` correctly? I'm not quite sure about the `egMapsLeafletLayer` option, should I remove it?

### DIFF
--- a/Maps_Settings.php
+++ b/Maps_Settings.php
@@ -318,6 +318,8 @@
 		// used when the user does not provide one.
 		$GLOBALS['egMapsLeafletLayer'] = 'OpenStreetMap';
 
+		$GLOBALS['egMapsLeafletLayers'] = ['OpenStreetMap'];
+
 		$GLOBALS['egMapsLeafletOverlayLayers'] = [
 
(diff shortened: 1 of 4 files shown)


PeterTheOne:
conflicts, see: #394

PR #394, opened 2017-11-16
leaflet: allow for multiple layers with control, fix #387

### LINKED ISSUE #387: Feature: Layer Groups and Layers Control for leaflet

PeterTheOne:
fixes #387 // cc @krabina

Adds support for multiple layers with leaflet. Displays layer controls like [this leaflet example](http://leafletjs.com/examples/layers-control/).

Only displays `L.control.layers` when there is more than one baselayer, control doesn't show up if there is for example a baselayer and an overlaylayer. overlaylayer are nevertheless included in the control. Should the default behavior be different? I could display the control once there is at least another baselayer or an overlaylayer. And should there be an extra option, like `forceLayerControl` (makes sense for the current default), `hideLayerControl` (usefull for the other default I just described)?

Should I rename `maps-leaflet-par-layer`? Is *"The layer that will be shown when the map loads"* still a good description?

@JeroenDeDauw Did I do the aliasing of `layer` / `layers` correctly? I'm not quite sure about the `egMapsLeafletLayer` option, should I remove it?

From my side it is ready to be merged, more options can also be added later on.


includes/Maps_DisplayMapRenderer.php
```diff
@@ -273,13 +273,15 @@ public static function getLayerDependencies( $service, $params ) {
...
-				&& array_key_exists( $layerName, $egMapsLeafletLayerDependencies ) ) {
-				$layerDependencies[] = '<script src="' . $egMapsLeafletLayerDependencies[$layerName] .
-					$egMapsLeafletLayersApiKeys[$layerName] . '"></script>';
+			$layerNames = $params['layers'];
+			foreach ( $layerNames as $layerName ) {
+				if ( array_key_exists( $layerName, $egMapsLeafletAvailableLayers )
+					&& $egMapsLeafletAvailableLayers[$layerName]
+					&& array_key_exists( $layerName, $egMapsLeafletLayersApiKeys )
+					&& array_key_exists( $layerName, $egMapsLeafletLayerDependencies ) ) {
+					$layerDependencies[] = '<script src="' . $egMapsLeafletLayerDependencies[$layerName] .
```
JeroenDeDauw: This HTML building is scary: no escaping

JeroenDeDauw:
> Should I rename maps-leaflet-par-layer? Is "The layer that will be shown when the map loads" still a good description?

Rename yes, else we'll get a `maps-leaflet-par-layers` message key shown on Special:Ask and such places

> Did I do the aliasing of layer / layers correctly?

I would expect the aliases to be an array but one element might work as well, knowing how I wrote code many years ago... Please do at least a manual test to see if it works: just use the alias and see if it gets recognized

> I'm not quite sure about the egMapsLeafletLayer option, should I remove it?

I would deprecate it (else it is a breaking change) and have the new one be 

```php
$GLOBALS['egMapsLeafletLayers'] = [ $GLOBALS['egMapsLeafletLayer'] ];
```

includes/Maps_DisplayMapRenderer.php
JeroenDeDauw: Does not need to be fixed in this PR, though cool if you did

PeterTheOne:
aliases work.

PeterTheOne:
I guess we could also merge `maps-leaflet-par-layers` and `maps-openlayers-par-layers` (maybe even `maps-googlemaps3-par-layers`).

JeroenDeDauw:
Could be... not sure these are necessarily the same though

krabina:
@JeroenDeDauw do you think you could make that into a 5.0.3 release?

JeroenDeDauw:
https://github.com/JeroenDeDauw/Maps/pull/395

PR #396, opened 2017-11-28
Update jquery.googlemap.js

Henryfunk:
The code contained a handful of unnecessary eval calls used to compute the names of properties. I have replaced these by bracket notation so that for instance: 

"mapOptions.zoomControlOptions = { style:eval('google.maps.ZoomControlStyle.' + options.zoomstyle) };"

becomes

"mapOptions.zoomControlOptions = { style:google.maps.ZoomControlStyle[ options.zoomstyle ] };" . 

I am not convinced eval should ALWAYS be avoided, but here it clearly is unnecessary. Using brackets is the standard solution for properties whose names aren't known at design time.

### DIFF
--- a/includes/services/GoogleMaps3/jquery.googlemap.js
+++ b/includes/services/GoogleMaps3/jquery.googlemap.js
@@ -471,7 +471,7 @@
 		this.initializeMap = function () {
 			var mapOptions = {
 				disableDefaultUI:true,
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Thank you!

Henryfunk:
You're welcome, and thank you both for all your good work on matters MediaWiki.

PR #405, opened 2018-01-22
Allow to configure geocoder cache expiry

TK-999:
**Problem**
The Google geocoder has a hard limit of 2500 requests / day. Depending on volume, administrators might want to tune the cache TTL of cached geocoded addresses, to avoid hitting the rate limit of this service.

**Solution**
Make the cache TTL of cached geocoded addresses optionally configurable, with the default value being the current hardcoded TTL (1 day).

### DIFF
--- a/Maps_Settings.php
+++ b/Maps_Settings.php
@@ -63,6 +63,9 @@
 
 	// Boolean. Sets if geocoded addresses should be stored in a cache.
 	$GLOBALS['egMapsEnableGeoCache'] = true;
+	// Integer. If egMapsEnableGeoCache is true, determines the TTL of cached geocoded addresses.
+	// Default value: 1 day.
+	$GLOBALS['egMapsGeoCacheTtl'] = BagOStuff::TTL_DAY;
 
 
 // Coordinate configuration
--- a/src/Geocoders/CachingGeocoder.php
+++ b/src/Geocoders/CachingGeocoder.php
@@ -16,10 +16,12 @@ class CachingGeocoder implements Geocoder {
 
 	private $geocoder;
 	private $cache;
+	private $cacheTtl;
 
-	public function __construct( Geocoder $geocoder, BagOStuff $cache ) {
+	public function __construct( Geocoder $geocoder, BagOStuff $cache, int $cacheTtl ) {
 		$this->geocoder = $geocoder;
 		$this->cache = $cache;
+		$this->cacheTtl = $cacheTtl;
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


JeroenDeDauw:
Thanks! Nicely implemented :)

kghbln:
I guess these are the possible options to `$egMapsGeoCacheTtl`:

* `BagOStuff::TTL_MINUTE` = 60
* `BagOStuff::TTL_HOUR` = 3600
* `BagOStuff::TTL_DAY` = 86400
* `BagOStuff::TTL_MONTH` = 2592000
* `BagOStuff::TTL_WEEK` = 604800
* `BagOStuff::TTL_YEAR` = 31536000
* `BagOStuff::TTL_INDEFINITE` = 0

TK-999:
Yeah, but they are really just useful shorthands. Any nonnegative integer should work fine there :)

kghbln:
> Yeah, but they are really just useful shorthands. Any nonnegative integer should work fine there :)

It worked since I wanted to trigger more info. :) So something like what we are seeing on the right of the equal sign is cool too like e.g. `$egMapsGeoCacheTtl = 86400;` or something like `$egMapsGeoCacheTtl = 259200;` (three days).

JeroenDeDauw:
I would not put `86400` since it is hard to read. Instead do `60 * 60 * 24` or something like that. Or perhaps even use the constants and do `BagOStuff::TTL_DAY * 3`.

I was also wondering about what stuff is allowed. Would be cool if you could add some info (to the point... not a whole book) to the setting.

TK-999:
Yeah there are lots of varieties such as `60 * 60 * 24; // 3 days` - since the whole thing ultimately depends on how `BagOStuff` interprets the number, which would be seconds in this case. I'll add some docs tomorrow.

Actually TIL that `BagOStuff` even accepts a Unix timestamp as TTL value, so technically a timestamp could also be valid for this configuration setting if someone wanted to set an arbitrary "doomsday" date when all cached items suddenly expire at once 💣 – although the usefulness of that seems debatable.

kghbln:
MediaWiki itself seems undecided on how to do it since I e.g. found all of this in "DefaultSettings.php":
- `$wgNewPasswordExpiry = 3600 * 24 * 7;`
- `$wgUserEmailConfirmationTokenExpiry = 7 * 24 * 60 * 60;`
- `$wgParserCacheExpireTime = 86400;`

I'd personally opt for e.g. 

`$egMapsGeoCacheTtl = 60 * 60 * 24 * 3;` for three days since we are talking about seconds. We know that there is a much bigger variety for sysadmins but we should also keep it simple.

PR #406, opened 2018-01-22
Change signature of MapsMappingService::addDependencies

TK-999:
It seems this method is only called in https://github.com/JeroenDeDauw/Maps/blob/master/includes/Maps_DisplayMapRenderer.php#L51 so it only ever gets called with a `Parser` instance.

I'm not sure how public this API is but none of the extensions listed [here](https://www.mediawiki.org/wiki/Extension:Maps#See_also) seem to be using it and historical precedent such as #215 only refers to Semantic Maps as a potential blocker, but SM is now assimilated.


JeroenDeDauw:
That might be from the pre-resourceloader (MediaWiki 1.17) days :D

PR #407, opened 2018-01-23
Leaflet: Add parameter to disable mouse scrolling (fix #316)

### LINKED ISSUE #316: can't disable mouse scrolling: Scrollwheel zooms map in and out when scrolling through the page
### Setup
https://wiki.sha2017.org/w/Special:Version

### Issue

Scrolling through a page using the mouse zooms the map when hovering over it. Would like to disable mouse scrolling and just use the control.

Example:
https://www.semantic-mediawiki.org/wiki/Maps_examples/OpenLayers_basic_map

Go here and and use your mouse wheel to scroll through the page.

Additionally: the static option doesn't work to fix this: when there are markers with large text-balloons, the position of the map is changed to show the text balloon after clicking: resulting you cannot use the mouse or anything else to show the original map except reloading the page. This makes the static option unusable.

hidrarga:
Should fix #316

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -50,7 +50,7 @@
 	"maps-coordinates-par-location": "The coordinates you want to format.",
 	"maps-coordinates-par-format": "The target format for the coordinates.",
 	"maps-coordinates-par-directional": "Indicates if the coordinates should be outputted directional or not.",
-	"maps-coordinates-par-scrollwheelzoom": "Indicates if mouse scrolling should be enabled or not.",
+	"maps-par-scrollwheelzoom": "Indicates if mouse scrolling should be enabled or not.",
 	"maps-distance-par-distance": "The distance to convert to its equivalent with a specified unit.",
 	"maps-distance-par-decimals": "The maximum number of fractional digits to use in the resulting value.",
 	"maps-distance-par-unit": "The unit to output the distance in.",
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -50,6 +50,7 @@
 	"maps-coordinates-par-location": "The coordinates you want to format.",
 	"maps-coordinates-par-format": "The target format for the coordinates.",
 	"maps-coordinates-par-directional": "Indicates if the coordinates should be outputted directional or not.",
+	"maps-coordinates-par-scrollwheelzoom": "Indicates if mouse scrolling should be enabled or not.",
 	"maps-distance-par-distance": "The distance to convert to its equivalent with a specified unit.",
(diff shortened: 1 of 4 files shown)


includes/Maps_Mapper.php
```diff
@@ -193,6 +193,11 @@ private static function getEvenMawrCommonParameters() {
 			'default' => false,
 		];
 
+		$params['scrollwheelzoom'] = [
+			'type' => 'boolean',
+			'default' => true,
+		];
```
JeroenDeDauw: Please add this to the leaflet specific parameters if it is only supported for that mapping service

i18n/en.json
```diff
@@ -50,6 +50,7 @@
 	"maps-coordinates-par-location": "The coordinates you want to format.",
 	"maps-coordinates-par-format": "The target format for the coordinates.",
 	"maps-coordinates-par-directional": "Indicates if the coordinates should be outputted directional or not.",
+	"maps-coordinates-par-scrollwheelzoom": "Indicates if mouse scrolling should be enabled or not.",
```
hidrarga: Let me know if that sentence is good enough ;)

includes/services/Leaflet/Maps_Leaflet.php
```diff
@@ -67,6 +67,11 @@ public function addParameterInfo( array &$params ) {
 			'message' => 'maps-par-enable-fullscreen',
 		];
 
+		$params['scrollwheelzoom'] = [
+			'type' => 'boolean',
+			'default' => true,
+		];
```
JeroenDeDauw: 'message' => 'maps-par-scrollwheelzoom',

i18n/en.json
JeroenDeDauw: Yeah but the key is now

'maps-par-scrollwheelzoom'

This is not a parameter for the coordinates parser function :)

includes/services/Leaflet/Maps_Leaflet.php
hidrarga: Oops, sorry! Didn't pay enough attention here..

JeroenDeDauw:
Thanks! Feel free to also add a line to the release notes

PeterTheOne:
qqq.json change still seems to be missing..

kghbln:
@PeterTheOne Will add this via twn.

kghbln:
https://translatewiki.net/wiki/MediaWiki:Maps-par-scrollwheelzoom/qqq

PR #408, opened 2018-01-30
5.2.0 release

JeroenDeDauw:


PeterTheOne:
#409 would make the `scrollwheelzoom` change more consistent.

JeroenDeDauw:
If I understand this correctly it does not actually change the default behavior, or the behavior added in this release. Hence I'll keep this for the next one.

PR #420, opened 2018-05-05
Re-arrange order of available mapping services

kghbln:
Since `egMapsDefaultServices` defaults to the first service defined with `egMapsAvailableServices` if no service is enabled or an invalid service was provided by the user we should have a service here that is freely usable and does not require extra setup to be functional.


JeroenDeDauw:
Thanks!

PR #424, opened 2018-05-17
Make tests compatible with PHPUnit 4 and 6

legoktm:
Use the PHPUnit4And6Compat trait from MediaWiki core to make PHPUnit 4
methods work even when being run with PHPUnit 6. If it's being run with
an older MediaWiki core version, define a empty trait in the bootstrap
to avoid fatal errors. Since the lack of the trait means we're running
with PHPUnit 4, everything should work out.

### DIFF
--- a/tests/Integration/parsers/DistanceParserTest.php
+++ b/tests/Integration/parsers/DistanceParserTest.php
@@ -4,6 +4,7 @@ namespace Maps\Test;
 
 use Maps\DistanceParser;
 use PHPUnit\Framework\TestCase;
+use PHPUnit4And6Compat;
 use ValueParsers\ParseException;
 
 /**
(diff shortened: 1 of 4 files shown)


JeroenDeDauw:
Thanks!

JeroenDeDauw:
Aaahhh. I should have checked the test output first.

image

legoktm:
Like I said on wikitech-l, that's a pre-existing problem on master, before my PR.

JeroenDeDauw:
Oh, had not seen your mail yet.

Indeed on current master they are not running, which is the issue I want to fix. This issue has been there since I switched the tests to PHPUnit 6.x+ format. Before that they ran fine https://travis-ci.org/JeroenDeDauw/Maps/builds/363977529

PR #426, opened 2018-05-17
Make tests compatible with PHPUnit 4 and 6 (try #2)

legoktm:
Reverts JeroenDeDauw/Maps#425, reinstates #424.


legoktm:
And that should do it!

PR #431, opened 2018-05-22
Remove use of deprecated wgStyleVersion

Krinkle:
I couldn't find any use of egMapsStyleVersion, but to keep this commit simple,
I kept that variable and only removed the use of wgStyleVersion. Also, in case
something depends on its format I've substituted it with the number 1.

This way any use of it will still work and still invalidate cache if the Maps_VERSION
constant is changed.

Ref <https://phabricator.wikimedia.org/T181318>


JeroenDeDauw:
Thanks! Looks like egMapsStyleVersion is indeed unused.

PR #436, opened 2018-06-10
change providers options and global options order

gcebelieu:
In leaflet-providers.js, layerOptions and global options are merged in a wrong order so that OSM attribution allways appear as the layer attribution whatever provider is used.

This PR simply inverts the options order. As a consequence, the right layer attribution appears on the map.

scotland-att


JeroenDeDauw:
Thank you!

PR #437, opened 2018-06-10
Adding some Geoportail France Maps

gcebelieu:
This PR adds the possibility to display French Geoportal Maps to Maps extension with leaflet provider.
Four maps types are added : 

- GeoportailFrance, 
- GeoportailFrance.orthos,
- GeoportailFrance.parcels,
- GeoportailFrance.ign_maps

Note that Geoportal Maps need an access key to be displayed. This PR uses a default public key that can be used as is but comes with no guarantee. 

Exemples : 

```
{{#display_map: 
 |layer=GeoportailFrance
 |service=leaflet
}}
```

gpmaps

```
{{#display_map: 
 |layer=GeoportailFrance.orthos
 |service=leaflet
}}

```

gporthos

```
{{#display_map: 
 |layer=GeoportailFrance.parcels
 |service=leaflet
}}

```

gpparcels

```
{{#display_map: 
 |layer=GeoportailFrance.ign_maps
 |service=leaflet
}}

```

ignmaps

### DIFF
--- a/includes/services/Leaflet/leaflet-providers/leaflet-providers.js
+++ b/includes/services/Leaflet/leaflet-providers/leaflet-providers.js
@@ -83,7 +83,7 @@
 			provider.options.attribution = attributionReplacer(provider.options.attribution);
 
 			// Compute final options combining provider options with any user overrides
-			var layerOpts = L.Util.extend({}, provider.options, options);
+			var layerOpts = L.Util.extend({}, options, provider.options);
 			L.TileLayer.prototype.initialize.call(this, provider.url, layerOpts);
 		}
 	});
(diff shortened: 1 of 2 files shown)


includes/services/Leaflet/leaflet-providers/leaflet-providers.js
```diff
@@ -628,7 +628,38 @@
...
+        GeoportailFrance: {
+			url: '//wxs.ign.fr/{apikey}/geoportail/wmts?REQUEST=GetTile&SERVICE=WMTS&VERSION=1.0.0&STYLE=normal&TILEMATRIXSET=PM&FORMAT={format}&LAYER={variant}&TILEMATRIX={z}&TILEROW={y}&TILECOL={x}',
+			options: {
+				attribution: '<a target="_blank" href="https://www.geoportail.gouv.fr/">Geoportail France</a>',
+				bounds: [[-75, -180], [81, 180]],
+				minZoom: 2,
+				maxZoom: 18,
+                // Get your own geoportail apikey here : http://professionnels.ign.fr/ign/contrats/ 
+                // NB : 'choisirgeoportail' is a demonstration key that comes with no guarantee
+                apikey: 'choisirgeoportail',
```
JeroenDeDauw: Requiring people to edit this JS file to change the api key is not ideal. But it is better than nothing :)

includes/services/Leaflet/leaflet-providers/leaflet-providers.js
```diff
@@ -628,7 +628,38 @@
...
+				ign_maps: 'GEOGRAPHICALGRIDSYSTEMS.MAPS',
+				maps: 'GEOGRAPHICALGRIDSYSTEMS.MAPS.SCAN-EXPRESS.STANDARD',
+				orthos: {
+					options: {
+						maxZoom: 19,
+						variant: 'ORTHOIMAGERY.ORTHOPHOTOS'
+					}
+				}
+			}
+        }
```
JeroenDeDauw: The indent here is rather messed up. I can fix that easily enough after merge so no big deal, but beware that normally proper indent is required.

includes/services/Leaflet/leaflet-providers/leaflet-providers.js
gcebelieu: I agree with you. Setting it in Maps_Settings.php would have been the right place for this (as for Mapbox key for instance), I made some unsuccesful tries to do it that way and temporaly gave up.

If you have some hints to do this that way, I will change this in a new PR.

includes/services/Leaflet/leaflet-providers/leaflet-providers.js
gcebelieu: Sorry for this.

includes/services/Leaflet/leaflet-providers/leaflet-providers.js
JeroenDeDauw: https://github.com/JeroenDeDauw/Maps/search?q=mw.config.get&unscoped_q=mw.config.get

gcebelieu: Thanks, I'll cast an eye on it.

JeroenDeDauw:
@gcebelieu  I did not realize this when merging the PR: you modified an included library. That means when the library is upgraded, we will lose your changes unless we re-apply them. So it is really better to define the layers elsewhere. That might also make it easier to access the settings.

gcebelieu:
@JeroenDeDauw you're right ! I didn't realize either that leaflet-providers was an external project.

Ideally I may try to push this PR on that project and think back settings stuff here or there.

JeroenDeDauw:
Right. Seems like time anyway to update the copy we have here as it is two years old.

PeterTheOne:
https://github.com/leaflet-extras/leaflet-providers

gcebelieu:
Hi @PeterTheOne, @JeroenDeDauw, 
It has been merged into leaflet-providers master.

JeroenDeDauw:
Has it also been released? Then I can just upgrade the library here to its latest version.

gcebelieu:
Yes it has just been released :

https://github.com/leaflet-extras/leaflet-providers/releases/tag/1.3.0

PR #440, opened 2018-07-01
Update Leaflet from 1.1.0 to 1.3.1

JeroenDeDauw:


JeroenDeDauw:
@gcebelieu Would be awesome if you could check if things still work fine after these changes, in particular the new stuff you added.

includes/services/Leaflet/leaflet-providers/leaflet-providers.js
```diff
@@ -640,17 +742,19 @@
...
 				parcels: {
 					options : {
 						variant: 'CADASTRALPARCELS.PARCELS',
 						maxZoom: 20,
+						style : 'bdparcellaire',
 						format: 'image/png'
 					}
 				},
-				ign_maps: 'GEOGRAPHICALGRIDSYSTEMS.MAPS',
+				ignMaps: 'GEOGRAPHICALGRIDSYSTEMS.MAPS',
```
gcebelieu: @JeroenDeDauw this change should be also made in Maps_Settings.php#379 : 

`'GeoportailFrance.ign_maps' => true,`

should be replaced by : 

`'GeoportailFrance.ignMaps' => true,`

PR #456, opened 2018-07-27
Extension.json and basic GeoJson pages

JeroenDeDauw:
WIP - do not merge


mwjames:
> +* Removed `Maps_VERSION` and `SM_VERSION` constants

Please make sure you provide an alternative since `SM_VERSION` is
actively used in SMW core.

Cheers

JeroenDeDauw:
Damn... then I won't remove these and will need to find a way of keeping them while having the extension.json stuff as well.

mwjames:
> Damn... then I won't remove these and will need to find a way of keeping
> them while having the extension.json stuff as well.

See [0] how we set `SMW_VERSION` while relying on `extension.json` to
give us the version number when the extension is being initialized.

Cheers

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/SemanticMediaWiki.php#L69-L74

JeroenDeDauw:
Great. Thanks for the link. Re-added the constants now.

PR #459, opened 2018-07-31
Add extra autoloading of entry point

JeroenDeDauw:
cause MW things...


JeroenDeDauw:
Actually this does not make sense

PR #473, opened 2018-09-07
Updates system messages

kghbln:
Refs: https://github.com/JeroenDeDauw/Validator/issues/31 in a way


kghbln:
@JeroenDeDauw I called it page name. Not sure if it should better be string, since we also have the corresponding verbose message explaining what is meant.

JeroenDeDauw:
Thanks!

Issue #493, opened 2018-11-16
Google Maps traffic and bicycle layers not working


acnetj:
Google Maps traffic and bicycle layers not showing.

Not working with 5.6 version. Was working under 5.0 version.

JeroenDeDauw:
Did you type those versions correctly? It worked in 5.0?

Also please provide wikitext for showing a map with these layers or link to the docu

acnetj:
Below is the code: https://www.transit.wiki/Bike_map

{{#compound_query:[[latlong::{{#var:latlong}} ({{#urldecode:{{#urlget:distance|2 mile}}}})]][[Category:Rail stations]];?latlong;icon=Rail20.png
    |[[latlong::{{#var:latlong}} ({{#urldecode:{{#urlget:distance|2 mile}}}})]][[Category:Bus stops]];?latlong;icon=Bus20.png
    |[[latlong::{{#var:latlong}} ({{#urldecode:{{#urlget:distance|2 mile}}}})]];?latlong;icon=images/f/f5/Mm_20_red.png;?Served by=Line;?Rname=Stop
    |limit=100|format=googlemaps|locations={{#var:latlong}}|center={{#var:latlong}}|width=80%|height=400|kml={{#urlget:kml}}|zoom=15|layers=bicycling}}

It is also not working on the Semantic Media Wiki's example: https://www.semantic-mediawiki.org/wiki/Maps_examples/Google_Maps_layer_map

JeroenDeDauw:
Confirmed. This was broken by https://github.com/JeroenDeDauw/Maps/commit/7c1f9686ced4a9b3fe8dcc9617adcc9912924a4c

JeroenDeDauw:
The fix for this issue is part of Maps 6.1.0 which was released today https://github.com/JeroenDeDauw/Maps/blob/master/RELEASE-NOTES.md#maps-610

PR #495, opened 2018-11-17
Fix Google Maps layer parameter

JeroenDeDauw:
Fixes https://github.com/JeroenDeDauw/Maps/issues/493


JeroenDeDauw:
I am not able to test this at present. Would be cool if someone could try it. Should be easy to copy over the change to Maps 5.x if need be.

Ping @acnetj @kghbln

PR #498, opened 2018-11-19
Update Maps_Settings.php

acnetj:
Add 'rotate' to egMapsGMaps3Controls to enable controls on 45 degree aerial view.

### DIFF
--- a/Maps_Settings.php
+++ b/Maps_Settings.php
@@ -206,7 +206,8 @@
 			'zoom',
 			'type',
 			'scale',
-			'streetview'
+		        'streetview',
+			'rotate'
 		];
 
 		// String. The default style for the type control.


JeroenDeDauw:
Thanks for the patch? Can you explain why this is a more reasonable default? Or provide a screenshot of a map with the control and the same map without the thing? (I cannot easily test this myself at the moment)

acnetj:
Below is the original code. Once you zoom in close enough it goes to 45 degree view that you cannot either go back to overhead view, or change angle.
image

This edit would add the rotate control like on this screen (see bottom left):
image

So you can chose to exit to go back to overhead view with the same zoom level:
image

Or change the angle to see the other side of the buildings:
image

These control are part of the default UI. Since the default UI is not activated by this extension it would have to be added in separately. The 'rotate' control is listed on the Maps_GoogleMaps3.php but it is not good unless it also appears on Maps_Settings.php.

PR #507, opened 2018-11-21
Move i18n out of root dir

JeroenDeDauw:
@Nikerabbit @siebrand Will this make anything blow up?

Also, is there a better way to name those files and perhaps a sensible name for a directory that holds both?


Nikerabbit:
Nothing will blow up.

JeroenDeDauw:
\o/

PR #541, opened 2019-04-11
Make table more compact

kghbln:


kghbln:
@JeroenDeDauw A squash and merge will be cool.

JeroenDeDauw:
Why this change?

kghbln:
It saves one ore two centimeters in width which is good on mobile devices I guess.

JeroenDeDauw:
I like the SMW newline, though to be honest I think the table was easier to understand without the other change

kghbln:
Hmm, I guess this is just a matter of habit.

PR #542, opened 2019-05-01
Remove pagelinktext param from KML format

JeroenDeDauw:
This has been broken since at least 2013.
Not clear to me what the desired behavior is, and if anyone
cares about having it.

### DIFF
--- a/RELEASE-NOTES.md
+++ b/RELEASE-NOTES.md
@@ -3,6 +3,10 @@ different releases and which versions of PHP and MediaWiki they support, see the
 [platform compatibility tables](INSTALL.md#platform-compatibility-and-release-status).
 
 
+## Maps 7.3.0
+
+* Removed long broken `pagelinktext` option from KML result format
+
 ## Maps 7.2.0
 
 Released on March 5th, 2019.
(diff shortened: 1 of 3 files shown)


JeroenDeDauw:
@mwjames I have the impression some SMW parameters are not supported or properly supported by the maps result formats based on what I saw here. Not sure what the desired behavior of those is though.

mwjames:
Hi,

> @mwjames I have the impression some SMW parameters are not supported or
> properly supported by the maps result formats based on what I saw here.

I have some difficulties to understand the hinted issue here, [0]
provides a couple of default parameters and anything beyond that is a
matter of how an individual result printer defines required parameters
and interprets them. The `pagelinktext` parameter is (or was) part of
Maps and is (or was) not related to any other SMW provide core
parameter.

> properly supported by the maps result formats based on what I saw here. Not
> sure what the desired behavior of those is though.

According to the description "semanticmaps-default-kml-pagelink":
"View page $1" therefore I would have assumed this provides a link to
the KML file (replacing $1) hereby allowing a user to view the content
of the KML file.

[0] https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Query/Processor/DefaultParamDefinition.php

Cheers

Issue #552, opened 2019-07-23
Class 'Maps\MediaWiki\Api\Geocode' not found


paladox:
### Setup

- Maps version: 7.3.1
- MW version: 1.33.0
- PHP version: 7.2.19
- SMW version (if applicable): none

### Issue

Hi, we are seeing this "/w/api.php   Error from line 200 of /srv/mediawiki/w/includes/api/ApiModuleManager.php: Class 'Maps\MediaWiki\Api\Geocode' not found"

JeroenDeDauw:
Thanks for reporting and attempting to fix! Will fix later

JeroenDeDauw:
@paladox what are you using the geocode API for?

paladox:
@JeroenDeDauw  I’m not using it, I saw the extension cause the api to throw an exception (e.g on /w/api.php) and also saw it in the logs for the other wikis.

JeroenDeDauw:
Alright. I will just remove it for now, as it was broken in at least one other way as well and I don't know anyone that uses it. Will re-add if users announce themselves on wikitech-l.

PR #554, opened 2019-07-24
Fix geocode API module

JeroenDeDauw:
Fixes https://github.com/JeroenDeDauw/Maps/issues/552


JeroenDeDauw:
@paladox this probably fixed the issue. Will write a regression test later today to confirm

JeroenDeDauw:
Note: only tests on master are failing (for unknown reason)

PR #573, opened 2019-09-29
Add PLURAL in maps-json-editor-edit-removed-shapes

jhsoby:


JeroenDeDauw:
Thank you!

Unfortunately this does not work. It's showing up in the revision history as `{{PLURAL:5|shape|shapes}}`. I do not know why. I asked on IRC yesterday and got no reply, so decided to drop the PLURAL for now. Guess I'll have to ask on the mailing list.

JeroenDeDauw:
See https://lists.wikimedia.org/pipermail/wikitech-l/2019-September/092599.html

jhsoby:
Oh, that's weird. I even checked how the message is used in the code, and assumed that mw.message would handle it correctly. Could the numParams thing from [here](https://www.mediawiki.org/wiki/Manual:Messages_API#mw.msg) have something to do with it?

jhsoby:
It seems like you need jqueryMsg for this to work, see https://www.mediawiki.org/wiki/Manual:Messages_API#Using_messages_in_JavaScript

JeroenDeDauw:
Ah yes. Looks like you are right about jqueryMsg. Brian Wolff  also mentioned it on the mailing list.

Feel free to add that change to the PR here. The map.leaflet.editor module is defined in extension.json.

I can get to reviewing this or doing it myself tomorrow.

jhsoby:
Done. Hope it works, I don't know how to actually test it. ;-)

JeroenDeDauw:
image

Hurray! Thanks a lot!

PR #582, opened 2019-10-12
Adds support for dark mode in Leaflet maps

vedmaka:
Allows to set different provider for users with dark appearance enabled on their system:

- adds `window.sm.isUserUsesDarkMode` function to detect dark theme
- adds `egMapsLeafletLayerDark` and `egMapsLeafletLayersDark` to specify providers for dark mode (defaults to `CartoDB.DarkMatter`)

### DIFF
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -156,6 +156,11 @@ return [
 	'egMapsLeafletLayer' => 'OpenStreetMap',
 	'egMapsLeafletLayers' => [ 'OpenStreetMap' ],
 
+	// String. The default layer for Leaflet to be used in dark theme mode.
+	// This value will only be used when the user does not provide one.
+	'egMapsLeafletLayerDark' => 'CartoDB.DarkMatter',
+	'egMapsLeafletLayersDark' => [ 'CartoDB.DarkMatter' ],
+
 	'egMapsLeafletOverlayLayers' => [],
 
 	// The definitions for the layers that should be available for the user.
--- a/DefaultSettings.php
+++ b/DefaultSettings.php
@@ -156,9 +156,8 @@ return [
 	'egMapsLeafletLayer' => 'OpenStreetMap',
 	'egMapsLeafletLayers' => [ 'OpenStreetMap' ],
 
-	// String. The default layer for Leaflet to be used in dark theme mode.
-	// This value will only be used when the user does not provide one.
-	'egMapsLeafletLayerDark' => 'CartoDB.DarkMatter',
+	// Array of strings. The default layers for Leaflet to be used in dark theme mode.
+	// This will override any user parameter set for "layers" or "layer".
 	'egMapsLeafletLayersDark' => [ 'CartoDB.DarkMatter' ],
 
 	'egMapsLeafletOverlayLayers' => [],
--- a/resources/leaflet/jquery.leaflet.js
+++ b/resources/leaflet/jquery.leaflet.js
@@ -5,7 +5,7 @@
  * @author Pavel Astakhov < pastakhov@yandex.ru >
  * @author Peter Grassberger < petertheone@gmail.com >
  */
-(function ($, mw, L, sm) {
+(function ($, mw, L) {
 	$.fn.leafletmaps = function ( options ) {
 		var _this = this;
 		this.map = null;
@@ -13,7 +13,6 @@
 		this.markers = [];
 		this.markercluster = null;
 		var apikeys = mw.config.get('egMapsLeafletLayersApiKeys');
-		var layersDark = mw.config.get('egMapsLeafletLayersDark');
 
 		/**
 		 * array point of all map elements (markers, lines, polygons, etc.)
@@ -299,6 +298,14 @@
 			return window.matchMedia( '(prefers-color-scheme: dark)' ).matches;
 		};
 
+		this.getLayers = function () {
+			if ( this.isUserUsesDarkMode() ) {
+				return mw.config.get('egMapsLeafletLayersDark');
+			}
+
+			return options.layers;
+		};
+
 		this.setup = function () {
 
 			var mapOptions = {};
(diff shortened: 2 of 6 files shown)


resources/leaflet/jquery.leaflet.js
```diff
@@ -336,7 +336,11 @@
 			this.bindClickTarget();
 
 			var layers = {};
-			$.each(options.layers.reverse(), function(index, layerName) {
+			var optionLayers = options.layers;
+			if ( sm.isUserUsesDarkMode() ) {
```
JeroenDeDauw: I'm unsure why you put this function in semanticMaps.js. Could you move it into this file?

src/LeafletService.php
```diff
@@ -50,6 +50,15 @@ public function getParameterInfo(): array {
 			'islist' => true,
 		];
 
+		$params['layers_dark'] = [
+			'aliases' => 'layer',
```
JeroenDeDauw: This alias is already defined and is not appropriate for this parameter

src/LeafletService.php
```diff
@@ -50,6 +50,15 @@ public function getParameterInfo(): array {
 			'islist' => true,
 		];
 
+		$params['layers_dark'] = [
```
JeroenDeDauw: To match the naming scheme of the other parameters this should be `darklayers`, perhaps with an added `dark_layers` alias.

DefaultSettings.php
```diff
@@ -156,6 +156,11 @@
 	'egMapsLeafletLayer' => 'OpenStreetMap',
 	'egMapsLeafletLayers' => [ 'OpenStreetMap' ],
 
+	// String. The default layer for Leaflet to be used in dark theme mode.
+	// This value will only be used when the user does not provide one.
+	'egMapsLeafletLayerDark' => 'CartoDB.DarkMatter',
+	'egMapsLeafletLayersDark' => [ 'CartoDB.DarkMatter' ],
```
JeroenDeDauw: Only the plural one is used in your new code as far as I can tell

resources/leaflet/jquery.leaflet.js
vedmaka: 😁 sure, actually, it was here but the last moment I decided it would be a better fit for `semanticMaps.js` - my idea was like it's something global and maybe could be used by other map services scripts in future

JeroenDeDauw:
You've added a setting (used in LocalSettings.php) are a parameter (used in wikitext). My question is if the parameter is really needed. For instance if the reasoning here is "most things have a parameter so I'll add one for this as well", then I rather not add it to the extension.

vedmaka:
Oh, wait a second.., I think it's a good question actually - my idea was that similarly to `egMapsLeafletLayers` the `egMapsLeafletLayersDark` allows to change the default provider(s) for dark appearance and the former `layers_dark` parameter is to be used as an instance override. So the question is - do we really need a per-instance override? Probably we do, because if there will be a map with custom layers set a user may want to change providers for dark version too. I am actually not very sure about this now, so I think it's your call.

JeroenDeDauw:
More features is more maintenance and more things to wade through for users. So I do not want to add something because "it might be useful to someone at some point". The dark layer stuff on its own seems quite fringe to me as it is. Hence I rather not have the per-instance parameter. Then again, implementing this without having the parameter might add more complexity... 

What is this "dark theme" anyway? Is this is browser thing or a MediaWiki thing? How do I test your new feature?

JeroenDeDauw:
And why are you adding this feature? Is this something you personally need? A client of yours? Multiple people? Just you for now but you know of other people who'd want it?

vedmaka:
> More features is more maintenance and more things to wade through for users. So I do not want to add something because "it might be useful to someone at some point". ..Hence I rather not have the per-instance parameter.

I can't disagree, let it be so

> What is this "dark theme" anyway? Is this is browser thing or a MediaWiki thing? How do I test your new feature?

well, it's just a system dark theme, eg. on Mac and Win 10 you can pick between dark / light appearance and this is being reflected in the corresponding media query n browsers so, you know, to be able to display websites in dark for people who prefer dark theme. [1] [2]

> And why are you adding this feature? Is this something you personally need? A client of yours? Multiple people? Just you for now but you know of other people who'd want it?

I think it's both for a client and for myself, since client is running a Mediawiki site with dark theme supporting skin so all the maps instances are outstanding in dark mode. And for myself because I am using dark mode setting. Overall, since dark modes become more and more popular and there is already a spec drafted for it in I assume it could be nice to have it supported in some way.

1. https://webkit.org/blog/8840/dark-mode-support-in-webkit/
2. https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

resources/leaflet/jquery.leaflet.js
```diff
@@ -336,7 +341,11 @@
 			this.bindClickTarget();
 
 			var layers = {};
-			$.each(options.layers.reverse(), function(index, layerName) {
+			var optionLayers = options.layers;
+			if ( this.isUserUsesDarkMode() ) {
+				optionLayers = layersDark;
+			}
+			$.each( optionLayers.reverse(), function(index, layerName) {
```
JeroenDeDauw: You can avoid the state and mutation you are doing here by introducing a new function:

```js
$.each( getLayers().reverse(), function(index, layerName) {}

function getLayers() {
    if ( this.isUserUsesDarkMode() ) {
        return mw.config.get('egMapsLeafletLayersDark');
    }

   return options.layers;
}
```

resources/leaflet/jquery.leaflet.js
```diff
@@ -461,4 +470,4 @@
 		return this;
 
 	};
-})(jQuery, window.mediaWiki, L);
+})(jQuery, window.mediaWiki, L, window.sm);
```
JeroenDeDauw: This is not needed now

JeroenDeDauw:
Great! Thanks for explaining. Getting the config is also easier than I though, so this solution is definitely more maintainable.

Just some minor code quality stuff left. I'll take care of that myself next time I touch Maps if you rather not change it once again.

PR #593, opened 2019-12-23
Fail to set up SMW JsonScript

JeroenDeDauw:


JeroenDeDauw:
@mwjames I have no idea why this is not working. I copied over things from ModernTimeline.

Would be nice to have a simple way in PHP to get what the query result of a query is given some data in SMW. Without needing to do weird autoloading, dealing with a deep class hierarchy or a new testing DSL. Basically a more sane version of https://github.com/JeroenDeDauw/Maps/blob/7.15.0/tests/System/SemanticMW/MapQueryTest.php.

Is that already exposed in some way?

JeroenDeDauw:
Muh. Actually would have a lot less need for this if the result printer design was fixed (https://github.com/SemanticMediaWiki/SemanticMediaWiki/pull/4313).

PR #602, opened 2020-02-12
WIP store geojon data in SMW

JeroenDeDauw:


MapsRegistration.php
```diff
@@ -43,6 +43,11 @@ public static function onRegistration( array $credits ) {
 				);
 			}
 
+			// FIXME: does not work. Registration apparently too late.
+			if ( !$GLOBALS['egMapsDisableSmwIntegration'] && defined( 'SMW_VERSION' ) ) {
+				$GLOBALS['smwgNamespacesWithSemanticLinks'][NS_GEO_JSON] = true;
+			}
```
JeroenDeDauw: @mwjames I'm not sure how to add this namespace to smwgNamespacesWithSemanticLinks. Apparently doing this in `wgExtensionFunctions` does not work, which is surprising to me. My guess is that registration here is too late since something already initialized the SMW code that bootstraps itself from $GLOBALS. If I do this outside of wgExtensionFunctions then it is too early, since SMW is not loaded yet (assuming you first load Maps and then SMW).

This PR works as expected when I just stuff 420 (the value of NS_GEO_JSON) into the smwgNamespacesWithSemanticLinks array in SMWs DefaultSettings.php.

Is not being able to change SMW config from wgExtensionFunctions intentional? I suspect this used to work at some point.

JeroenDeDauw:
@mwjames another question: I noticed that the data stored with this code does not show up on Special:Browse if the job queue is not emptied. Is that expected? And if so, is there a way to avoid needing to run the queue for this data? (Ideally all the work is done during the page saving process.)

mwjames:
> smwgNamespacesWithSemanticLinks. Apparently doing this in wgExtensionFunctions does not work, which is surprising to me. My guess is that registration here is too late since something already initialized the SMW code that bootstraps itself from $GLOBALS. If I do this outside of wgExtensionFunctions then it is too early, since SMW is not loaded yet (assuming you first load Maps and then SMW).
> Is not being able to change SMW config from wgExtensionFunctions intentional? I suspect this used to work at some point.

See https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/4505#issuecomment-586564004.

JeroenDeDauw:
Thanks

PR #604, opened 2020-02-18
FileUrlFinder: fix marker images not being found in Shared repository

edwardspec:
Previously MediaWikiFileUrlFinder was checking for the presence of "File:Something.png" page in the local wiki, but this page may be nonexistent for images that are in Wikimedia Commons, etc.

Proper way is to query RepoGroup::findFile() to find the file (it will find files from shared repositories too).


JeroenDeDauw:
Thanks!

JeroenDeDauw:
https://github.com/JeroenDeDauw/Maps/commit/c0586fe88e7275d75104acf3c16b3110f3d84a8b

PR #614, opened 2020-04-12
Add cargo integration

JeroenDeDauw:
image


JeroenDeDauw:
@yaronkoren FYI. So far this only overrides the "map" format. "openlayers", "leaflet" and "googlemaps" have been left alone. This way you can still get the Cargo version if you really want. "map" will get you either a Leaflet or Google Maps map, depending on how you configured the Maps extension.

PR #615, opened 2020-04-15
Add Special:MyLanguage for better user experience

shirayuki:
Add Special:MyLanguage to MediaWiki.org URLs for better user experience


kghbln:
Thanks a lot!

PR #683, opened 2022-03-15
Update reference of Freenode to Libera

fredster33:


kghbln:
Failure unrelated to code change

Issue #694, opened 2022-05-12
i18n: The message "maps-geo-json-create-page-button" should be in content view language or user language instead of content language


WinstonSung:
### Setup

- Maps version: 7.20.1
- MW version: 1.33.3 (upgrading to 1.37.2; see announcements from Fandom)
- PHP version: 7.3.33 (fpm-fcgi)
- SMW version (if applicable): none

### Issue

Detailed description of the issue and a [stack trace](https://www.semantic-mediawiki.org/wiki/Help:Identifying_bugs) if applicable:

`/src/GeoJsonPages/GeoJsonNewPageUi.php` :
```
wfMessage( 'maps-geo-json-create-page-button' )->inContentLanguage()->text()
```

Steps to reproduce the observation (recommendation is to use the [sandbox](https://sandbox.semantic-mediawiki.org)):

* Go to any wiki using language with language variants
    * Example: https://ucptest.fandom.com/zh/wiki/ using language `zh` with `zh, zh-hans, zh-hant, zh-cn, zh-hk, zh-mo, zh-myy, zh-sg, zh-tw` variants.
* Select `zh-hant`, `zh-tw`, `zh-hk` or `zh-mo` as the variant ("**page view language**" aka "**user language variant**", see [Manual:Language](https://www.mediawiki.org/wiki/Special:MyLanguage/Manual:Language) )
    * Example: https://ucptest.fandom.com/zh/wiki/GeoJson:T?variant=zh-tw
* Current behavior: using `/zh` translation for the "Create this page" button
* Expected behavior: using `/zh-tw` translation for the "Create this page" button

Related files:
```
/psalm-baseline.xml
/src/GeoJsonPages/GeoJsonNewPageUi.php
```

WinstonSung:
Solutions:

- Using page view language: replace `->inContentLanguage()` , using `->inLanguage( $lang )` with page view language as the variable passed to the method instead.
- Using user language: simply remove `->inContentLanguage()` .

JeroenDeDauw:
Thanks.

If you are able and willing to provide a PR, that is definitely welcome.

WinstonSung:
I'm having issues in `wfMessage( 'maps-loading-map' )->inContentLanguage()->text()` in `/src/GeoJsonPages/GeoJsonMapPageUi.php` as the function doesn't privide any information about the page language.

Should I just use user interface language then?

CC @JeroenDeDauw

WinstonSung:
[...]

PR #695, opened 2022-05-13
Make content-language-bounded messages display in user interface language

WinstonSung:
Make content-language-bounded messages display in user interface language

This fix the issue that the system message "maps-geo-json-create-page-button", "maps-loading-map" cannot display in language variants ("page view language" aka "user language variant", see Manual:Language on MediaWiki.org )

Bug: #694


WinstonSung:
CC @JeroenDeDauw

JeroenDeDauw:
I'm not sure this is correct. Especially `maps-loading-map`, which definitely ends up in the parser cache.

@Nikerabbit

WinstonSung:
I was considering using "page view language" rather than just using "user interface language" in namespaces with wikitext content modal, but there's no "page language"-related variables passed in, causing the complexity to implement it.

WinstonSung:
It seems the CI already failed before this commit?

> 3b585b3 Localisation updates from https://translatewiki.net.
> https://github.com/ProfessionalWiki/Maps/runs/6404720045

Nikerabbit:
`wfMessage` is implicit global state. Ideal solution would be to pass the correct language somehow.

JeroenDeDauw:
> wfMessage is implicit global state. Ideal solution would be to pass the correct language somehow.

Yes sure. But is the change in this PR correct or not?

> It seems the CI already failed for "PHPUnit: MW master, PHP 8.1" before this commit?

Indeed, don't worry about that failure

Nikerabbit:
I can't tell because I am not familiar with the code. Most likely not though, if this is ending up in the parser cache without fragmenting it by language.

PR #703, opened 2022-07-06
Deprecate SkinTemplateNavigation in favor of SkinTemplateNavigation::Universal

Abijeet:
SkinTemplateNavigation was deprecated in 1.39.

SkinTemplateNavigation::Universal is available since 1.18


Abijeet:
I've not had a chance to test this but I think it should work. 

Please squash and merge

Issue #698, opened 2022-06-11
Fix Content deprecations on MW 38


JeroenDeDauw:
> PHP Deprecated:  Maps\GeoJsonPages\GeoJsonContent overrides fillParserOutput which was deprecated in MediaWiki 1.38. [Called from MediaWiki\Content\Renderer\ContentRenderer::getParserOutput in /home/runner/work/Maps/Maps/mediawiki/includes/content/Renderer/ContentRenderer.php at line 47] in /home/runner/work/Maps/Maps/mediawiki/includes/debug/MWDebug.php on line 381

> PHP Deprecated:  Use of AbstractContent::getParserOutput was deprecated in MediaWiki 1.38. [Called from ContentHandler::callDeprecatedContentGPO in /home/runner/work/Maps/Maps/mediawiki/includes/content/ContentHandler.php at line 1873] in /home/runner/work/Maps/Maps/mediawiki/includes/debug/MWDebug.php on line 381

JeroenDeDauw:
See also: https://github.com/ProfessionalWiki/Maps/issues/709#issue-1345350829

PR #723, opened 2023-01-31
Remove deprecated method Content::fillParserOutput()

samwilson:
Move the HTML-setting into the ContentHandler instead of the Content.

Bug: #698

### DIFF
--- a/src/GeoJsonPages/GeoJsonContentHandler.php
+++ b/src/GeoJsonPages/GeoJsonContentHandler.php
@@ -4,6 +4,12 @@ declare( strict_types = 1 );
 
 namespace Maps\GeoJsonPages;
 
+use Content;
+use Maps\MapsFactory;
+use Maps\Presentation\OutputFacade;
+use MediaWiki\Content\Renderer\ContentParseParams;
+use ParserOutput;
+
 class GeoJsonContentHandler extends \JsonContentHandler {
 
 	public function __construct( $modelId = GeoJsonContent::CONTENT_MODEL_ID ) {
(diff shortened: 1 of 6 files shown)


JeroenDeDauw:
What are the compat implications of this change?

samwilson:
> What are the compat implications of this change?

It bumps the required MW version to 1.38. How important is it to maintain backwards compatibility? I think if it's necessary, then probably a separate legacy class of Content will be needed, because I don't think it's possible to optionally override a method (and so avoid the deprecation).

JeroenDeDauw:
Then this will be for the next major release of Maps: version 11. Will hold off with merging into master for a little bit.

JeroenDeDauw:
You could remove the MW 35-MW37 test runs form GitHub actions in this PR

Would also help to have the tests run and pass against MW 39, though that might require fixing things out of scope of this PR, and thus better done seperately

samwilson:
I'm attempting to keep 1.35 compatibility. Feels like it's not too complicated... see what you think!

src/GeoJsonPages/GeoJsonContentHandler.php
```diff
@@ -4,18 +4,49 @@
...
 
 	protected function getContentClass(): string {
-		return GeoJsonContent::class;
+		return version_compare( MW_VERSION, '1.38', '<' ) ? GeoJsonLegacyContent::class : GeoJsonContent::class;
 	}
 
 	public function makeEmptyContent(): GeoJsonContent {
-		return new GeoJsonContent( GeoJsonContent::newEmptyContentString() );
+		$class = $this->getContentClass();
+		return new $class( $class::newEmptyContentString() );
```
JeroenDeDauw, line 25: Why is this change needed?

samwilson: It's to avoid having to repeat the version check here, to make sure it returns an object of the right class (either GeoJsonContent or GeoJsonLegacyContent). This way, it won't have to be changed when eventually the legacy content class is removed.

PR #729, opened 2023-02-17
Consistently use egMapsScriptPath for front-end asset paths

mszabo-wikia:
Ensure all front-end asset paths use `egMapsScriptPath` as the base rather than building an asset path themselves. Derive `egMapsScriptPath` from the `ExtensionAssetsPath` config variable, since deployments may use this config var to customize the asset path for e.g. multi-version MediaWiki deployments.

### DIFF
--- a/src/MapsHooks.php
+++ b/src/MapsHooks.php
@@ -62,7 +62,10 @@ final class MapsHooks {
 	 * @return boolean true in all cases
 	 */
 	public static function onMakeGlobalVariablesScript( array &$vars ) {
-		$vars['egMapsScriptPath'] = $GLOBALS['wgScriptPath'] . '/extensions/Maps/'; // TODO: wgExtensionDirectory?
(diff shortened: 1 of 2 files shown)


src/MapsHooks.php
```diff
@@ -62,7 +62,10 @@ public static function addToAdminLinks( ALTree &$admin_links_tree ) {
 	 * @return boolean true in all cases
 	 */
 	public static function onMakeGlobalVariablesScript( array &$vars ) {
-		$vars['egMapsScriptPath'] = $GLOBALS['wgScriptPath'] . '/extensions/Maps/'; // TODO: wgExtensionDirectory?
+		$config = MediaWikiServices::getInstance()->getMainConfig();
+		$extensionAssetsPath = $config->get( 'ExtensionAssetsPath' );
```
JeroenDeDauw, line 66: This var is not needed

Issue #720, opened 2023-01-21
How to apply Google API key restrictions?


georgjaehnig:
To my knowledge, I can only set 1 Google API key with `$egMapsGMaps3ApiKey`, which then is used both for

- gecoding
- map display.

Now I'd like to Key restrictions in Google, to prevent that other people use my key (and then generate quote on my bill). Google offers me among others to restrict by

- HTTP referrers (web sites)
- IP addresses (web servers, cron jobs, etc.)

<img width="563" alt="image" src="https://user-images.githubusercontent.com/1092628/213882175-81c02969-12de-494b-9ba4-5174054c2fb2.png">

These are radio buttons so it seems I can only pick one of them at the same time. 

- So when I pick HTTP referrers (and set it to my website), then the map can get displayed – but the geocoder, operating from my server cannot.
- And when I pick IP addresses (and set it to my server), then the geocoder can work – but the map display does not work anymore because my website visitors have many different IPs each.

So how can I meaningfully restrict my Google API key? I thought about creating 2 (one for map display, one for geocoder), but Maps seems to accept only one in the settings.

JeroenDeDauw:
Maps indeed only support one Google API key

georgjaehnig:
OK. So you don't restrict your API key at all on your wikis?

JeroenDeDauw:
This was implemented in 2009, and the API keys have undergone some changes since then. Personally, I am not using Google Maps anywhere, so I am not up to speed.

Your use case sound valid to me. Maps simply does not support it. That said, I do welcome PRs that add the feature.

georgjaehnig:
Alright. I could implement the sth. like this. The API key could still either be set as a string:

    $egMapsGMaps3ApiKey = 'MY_API_KEY';

or as an array:

    $egMapsGMaps3ApiKey = [
      'Geocoding API' => 'MY_GEOCODING_KEY'
      'Maps JavaScript API' => 'MY_MAPS_JAVASCRIPT_API_KEY'
    ];

Make sense?

JeroenDeDauw:
I would rather keep the setting always be a string and instead have a new `egMapsGoogleGeocodingApiKey` setting. That one could then (internally) default to `egMapsGMaps3ApiKey` if it is not set.

PR #730, opened 2023-02-26
720 add google geocoding api key

georgjaehnig:
Solving #720.

### DIFF
--- a/src/MapsSetup.php
+++ b/src/MapsSetup.php
@@ -48,7 +48,7 @@ class MapsSetup {
 			$GLOBALS['egMapsGMaps3ApiKey'] = $GLOBALS['egGoogleJsApiKey'];
 		}
 
-		if ( empty($GLOBALS['egMapsGoogleGeocodingApiKey'])  && array_key_exists( 'egMapsGMaps3ApiKey', $GLOBALS ) ) {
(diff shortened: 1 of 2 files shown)


src/MapsSetup.php
```diff
@@ -47,6 +47,10 @@ private function defaultSettings() {
 		if ( $GLOBALS['egMapsGMaps3ApiKey'] === '' && array_key_exists( 'egGoogleJsApiKey', $GLOBALS ) ) {
 			$GLOBALS['egMapsGMaps3ApiKey'] = $GLOBALS['egGoogleJsApiKey'];
 		}
+
+		if ( $GLOBALS['egMapsGoogleGeocodingApiKey'] === '' && array_key_exists( 'egMapsGMaps3ApiKey', $GLOBALS ) ) {
```
JeroenDeDauw, line 51: Is `egMapsGoogleGeocodingApiKey` guaranteed to exist? If not, this line will cause an issue

georgjaehnig: I see. How about replacing `$GLOBALS['egMapsGoogleGeocodingApiKey'] === ''` with `empty($GLOBALS['egMapsGoogleGeocodingApiKey'])`?

JeroenDeDauw: Sure

PR #731, opened 2023-02-27
Fix support for composer 2.2.1+

Universal-Omega:
Composer 2.2.1+ requires plugins to be allowed


JeroenDeDauw:
AFAIK this is unnecessary if you run composer inside of MW itself and include the extension via the composer merge extension (composer.local.json). While this change allows running composer in the extensions directory itself, doing so is unsafe and not something we have been actively supporting for a while. How did you run into this?

cc @malberts

PR #736, opened 2023-04-08
FIX(leaflet): Load only one base layer at a time

dexgs:
Previously, leaflet maps would initially load and display *all* available base layers at once. This caused performance issues on clients with slow connections or limited memory.

The function that adds base layers to the map has been changed so that it only initially adds the last available base layer. This solves the problem while making sure the initially visible base layer stays the same (previously, although all base layers were loaded only the last layer was shown on top and was visible to the user).

### DIFF
--- a/resources/leaflet/jquery.leaflet.js
+++ b/resources/leaflet/jquery.leaflet.js
@@ -280,10 +280,11 @@
 			return layers;
 		};
 
-		this.addBaseLayersToMap = function(layers) {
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
Can users still see the other layers?

JeroenDeDauw:
Can you check if there is an example of this issue on https://maps.extension.wiki, or otherwise create one? Then I can easily see the effect of your change by deploying it there.

PR #743, opened 2023-06-07
Remove usage of wgHooks global variable

Abijeet:
PHP Deprecated:  Accessing $wgHooks directly is deprecated, use 
HookContainer::getHandlers() or HookContainer::register() instead

Bug: https://phabricator.wikimedia.org/T332687

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,7 @@ jobs:
         run: |
           php tests/phpunit/phpunit.php -c extensions/Maps --coverage-clover coverage.xml
           bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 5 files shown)


JeroenDeDauw:
Cannot merge this without assurance from the tests, which are once again broken due to some upstream changes. Sigh

Abijeet:
> Cannot merge this without assurance from the tests, which are once again broken due to some upstream changes. Sigh
> 
> It's the allow-plugins thing again. I do not see how the config of Maps is different from https://github.com/ProfessionalWiki/WikibaseExport where the issue does not occur.

I'll try to fix this

JeroenDeDauw:
Thanks!

PR #769, opened 2024-02-20
Remove executable bit from Maps_Settings.php

samwilson:
No need for this file to be executable.


samwilson:
Found because I was running [minus-x](https://packagist.org/packages/mediawiki/minus-x) over my extensions directory. That check is usually part of `composer test` but I assume is left out of Maps intentionally so I didn't add it for this.

PR #776, opened 2024-03-20
Do not call $parser->clearState() when parser options exist

pastakhov:
The previous patch breaks widgets (and most likely other things) because it calls $parser->clearState() every time.


tosfos:
Referencing #773

PR #783, opened 2024-09-06
Replace DB_MASTER with DB_PRIMARY

osnard:
DB_MASTER was deprecated in MediaWiki 1.36 and removed in MediaWiki 1.43


osnard:
See https://github.com/SemanticMediaWiki/SemanticMediaWiki/issues/5634#issuecomment-2333836086

PR #784, opened 2024-09-18
User geolocation in Google Maps

### LINKED ISSUE #98: Get user's current coordinates from browser
It should be possible to get the current location of a user (the user's browser) via JavaScript:
https://www.mediawiki.org/wiki/API:Showing_nearby_wiki_information#How_to_get_geographic_coordinates

This can then be used e. g. in Semantic Maps queries, or also simply in Maps, centering a map around a current location or displaying the current location on a map.

http://semantic-mediawiki.org/wiki/Semantic_Maps_examples/Distance_query - here the distance query is done with a fixed coordinate, it would be great if we could do:
[[Has coordinates::{{CURRENTLOCATION}} (0.3 km)]]

stevygee:
- Adds parameter 'mylocation'
- When enabled, geolocation toggle button is shown in map UI
- Adds marker and accuracy circle to map
- Watches user position, re-centers map
- When dragged, position keeps updating, but map no longer follows
- Geolocation can be turned off
- Fixes AJAX markers not updating when changing center only

Fixes #98


src/GoogleMapsService.php
```diff
@@ -273,7 +280,7 @@ public function newMapId(): string {
 	}
 
 	public function getResourceModules( array $params ): array {
-		return [ 'ext.maps.googlemaps3', 'ext.maps.googlemaps3ajax' ];
+		return [ 'ext.maps.googlemaps3', 'ext.maps.googlemaps3ajax', 'ext.maps.mylocation' ];
```
JeroenDeDauw, line 283: If you always load the new js file anyway, you could put it in one of the existing modules.

JeroenDeDauw:
Do you have a demo of this somewhere? Or a short video?

Flexmaen:
> Do you have a demo of this somewhere? Or a short video?

Here: https://baugeschichte.at/Spezial:KU/ff9
Might mike more sense if you use it in your phone,  and with a fake location set to some point that has entries around.

src/GoogleMapsService.php
stevygee: Sure, I'm not very familiar how script loading works in MediaWiki, but I've added it to the `ext.maps.googlemaps3` module now.

stevygee:
Added: The user location setting is now remembered (local storage), so it remains active for all maps on a site.

Added two new parameters:
- **mylocationfollow**: When tracking user location, also follow user location by re-centering map on them (on/off).
- **mylocationzoom**: Level to zoom to when activating user location. Set at -1 to disable (default).

(both of these are disabled in the demo)

PeterTheOne:
Looks good to me :)

PR #799, opened 2025-02-20
Fix reference to SMW class aliases that were removed

paladox:
## Summary by CodeRabbit

- **Documentation**
	- Clarified inline comments to improve method reference accuracy.
- **Refactor**
	- Updated parameter type hints for enhanced consistency across the codebase.


paladox:
Just some minor comment updates if you want them @JeroenDeDauw

JeroenDeDauw:
Better to use imports rather than using the FQNs directly

PR #802, opened 2025-03-04
Use data-mw-maps-mapdata attribute for mapdata

alistair3149:
## Summary by CodeRabbit

- **Refactor**
  - Streamlined map loading by retrieving data directly from HTML attributes instead of parsing text content.
  - Consolidated map data embedding to simplify the integration process, enhancing efficiency and reliability in map displays.
- **Tests**
  - Introduced a new assertion method to enhance HTML string comparisons across multiple test cases, ensuring proper escaping to prevent HTML injection.


JeroenDeDauw:
This seems worth bumping the major version number for, because you need to invalidate the parser cache of your wiki due to the HTML change. Otherwise the JS will get the old HTML and not find the data.

JeroenDeDauw:
I'll bump the version

JeroenDeDauw:
See https://github.com/ProfessionalWiki/Maps/pull/803

PR #805, opened 2025-03-05
Prefix validator message keys

malberts:
For https://github.com/ProfessionalWiki/Maps/pull/801#issuecomment-2700109172

There are some other existing Maps-specific `validation-` message keys. Not sure if it is worth prefixing them to make everyting consistent?

## Summary by CodeRabbit

- **Refactor**
  - Standardized maps-related validation and error messages across all supported languages for clearer, more context-specific notifications. No functional changes were introduced, ensuring a more consistent user experience when interacting with map features.


malberts:
@Nikerabbit if there is no other way to resolve the message key conflict, would this be fine?

Nikerabbit:
This patch is likely the best way. We can add overrides on translatewiki side, but usually that's not the best option as anyone installing two things using conflicting message keys would have a lottery on which ones are shown.

malberts:
Turns out, this PR is not enough. Maps depends on https://github.com/JeroenDeDauw/ParamProcessor/, which in turn is generating the `validator-...` keys:
https://github.com/JeroenDeDauw/ParamProcessor/blob/0850dc2af72d31b8f645e0f87a44ff4b68583a64/src/ParamDefinition.php#L261-L262

malberts:
On the other hand, the idea is to abandon the Validator extension in the whole MW ecosystem, so the user shouldn't have it after #801.

The remaining extensions that still list it as a dependency are very likely abandoned.

PR #835, opened 2026-01-29
Update jeroen/file-fetcher-cache version constraint

malberts:
Depends on https://github.com/JeroenDeDauw/file-fetcher-cache/issues/19

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
 		"param-processor/param-processor": "^1.10",
 		"data-values/geo": "~4.0|~3.0",
 		"jeroen/file-fetcher": "~6.0|~5.0",
-		"jeroen/file-fetcher-cache": "~1.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -42,7 +42,7 @@
 		"param-processor/param-processor": "^1.10",
 		"data-values/geo": "~4.0|~3.0",
 		"jeroen/file-fetcher": "~6.0|~5.0",
-		"jeroen/file-fetcher-cache": "~1.0",
+		"jeroen/file-fetcher-cache": "~1.0|~2.0",
```
malberts: While Maps 12 still supports PHP 7.4, we need to allow both versions.

v1 still supports PHP 7.4.
v2 requires PHP 8.0, because the upstream dependency (psr/simple-cache) requires PHP 8.0.

PR #837, opened 2026-02-17
Fix "Call to undefined method MediaWiki\Parser\ParserOutput::getText()"

paladox:
Copies something similar to https://github.com/SemanticMediaWiki/SemanticMediaWiki/blob/master/src/Utils/ParserOutputHelper.php#L29

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of wikitext-to-HTML rendering to handle multiple parser outputs and missing body content.
  * Added safer error handling to avoid failures when body content is absent, improving page reliability.
  * Result: more consistent rendering across varied inputs.

* **Tests**
  * Updated tests to reflect multi-stage parser output handling and defensive content extraction.

### DIFF
--- a/tests/Util/TestFactory.php
+++ b/tests/Util/TestFactory.php
@@ -4,6 +4,7 @@ declare( strict_types = 1 );
 
 namespace Maps\Tests\Util;
 
+use LogicException;
 use MediaWiki\MediaWikiServices;
 use MediaWiki\Title\Title;
 use User;


paladox:
Fixes MW 1.45 support (well one issue)

src/Presentation/WikitextParser.php
```diff
@@ -20,11 +21,26 @@ public function wikitextToHtml( string $text ): string {
...
+		if ( method_exists( $parserOutput, 'getContentHolderText' ) ) {
+			try {
+				return $parserOutput->getContentHolderText();
+			} catch ( LogicException $e ) {
+				// Handle case where there is no body content
+				return '';
+			}
+		} elseif ( method_exists( $parserOutput, 'getRawText' ) ) {
+			return $parserOutput->getRawText();
+		} elseif ( method_exists( $parserOutput, 'getText' ) ) {
```
JeroenDeDauw: Why also a case for `getRawText`?

paladox: Removed, thanks, done.

PR #839, opened 2026-02-17
Replace deprecated wfExpandUrl()

lewiscawte:
This function was deprecated in 1.39 and hard-deprecated in 1.45. As such, it needs to be replaced by the UrlUtils class equivalent to avoid deprecation warnings, and the global function will likely be removed in the near future.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal URL handling implementation to use modern service-based architecture for improved maintainability and consistency with current framework standards.

### DIFF
--- a/src/GoogleMapsService.php
+++ b/src/GoogleMapsService.php
@@ -6,6 +6,7 @@ namespace Maps;
 
 use MediaWiki\Html\Html;
 use Maps\Map\MapData;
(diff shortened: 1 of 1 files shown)


JeroenDeDauw:
LGTM

Looks like there is an issue with the CI tho

malberts:
PHPUnit passes after CI update, although something in CI is still messed up after running tests.

PR #899, opened 2026-05-28
Escape Leaflet layer control labels and enforce the available-layers whitelist

JeroenDeDauw:
_Written by Claude Code, Opus 4.8. Result of iterative work with @JeroenDeDauw. Context: the Maps extension codebase (Leaflet service and layer control)._

The Leaflet layer control rendered base layer, overlay, and image layer names without escaping, and the `layers`/`overlays` parameters were not restricted to the configured available layers (ParamProcessor only emitted a non-fatal warning for values outside the whitelist).

This escapes layer and overlay names where they are used as Leaflet layer control labels (`addOverlays` and `addLayersAndOverlays`), and filters the `layers` and `overlays` parameters to the `egMapsLeafletAvailableLayers` / `egMapsLeafletAvailableOverlayLayers` whitelists in `LeafletService` so non-enabled values are dropped rather than passed through to the client.

**Behavior note:** this now enforces the available-layers whitelist that was previously only warned about. Wikis using custom layers must ensure they are listed in `egMapsLeafletAvailableLayers` / `egMapsLeafletAvailableOverlayLayers`.

Adds unit tests for the whitelist filtering and QUnit tests asserting the layer control labels are escaped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/resources/leaflet/jquery.leaflet.js
+++ b/resources/leaflet/jquery.leaflet.js
@@ -291,7 +291,7 @@
 			let overlays = {};
 
 			$.each(options.overlays, function(index, overlayName) {
-				overlays[overlayName] = new L.tileLayer.provider(overlayName).addTo(_this.map);
+				overlays[mw.html.escape(overlayName)] = new L.tileLayer.provider(overlayName).addTo(_this.map);
 			});
 
 			return overlays;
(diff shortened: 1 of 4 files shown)


SomeMWDev:
Thanks!

